PR #65, opened 2015-03-12
Provide ability to suppress automatic generation of keys

GrabYourPitchforks:
This required the introduction of a "fallback key" concept, otherwise it would have been too easy for developers to turn off automatic key generation and forget to turn it back on before deploying to production, blissfully unaware of the ticking time bomb that is key expiration.


src/Microsoft.AspNet.DataProtection/DataProtectionConfiguration.cs
```diff
@@ -114,6 +114,36 @@ public DataProtectionConfiguration ConfigureGlobalOptions([NotNull] Action<DataP
...
 
         /// <summary>
+        /// Configures the data protection system not to generate new keys automatically.
+        /// </summary>
+        /// <returns>The 'this' instance.</returns>
+        /// <remarks>
+        /// Calling this API corresponds to setting <see cref="KeyManagementOptions.AutoGenerateKeys"/>
+        /// to 'false'. See that property's documentation for more information.
+        /// </remarks>
+        public DataProtectionConfiguration DisableAutomaticKeyGeneration()
```
davidfowl: We don't really have this pattern anywhere else. A single method to configure individual properties of a single options model.

Issue #73, opened 2015-04-01
DataProtection throws NullReferenceException on IIS


ChengTian:
Run dnu published MusicStore on IIS gives:
![untitled](https://cloud.githubusercontent.com/assets/1383883/6932731/79b7587c-d7d2-11e4-82ce-a67f004484d8.png)

ChengTian:
@muratg @Praburaj

Praburaj:
Looking at this now.

Praburaj:
I see both x86 versions (clr and coreclr) fails with this error. Both X64 versions work fine. May be something wrt bitness. investigating this more.

Praburaj:
merged : https://github.com/aspnet/DataProtection/commit/33fa06a25c8823cd3f73185573831ed6f769fbd9

PR #75, opened 2015-04-01
Fixing dataprotection on IIS

Praburaj:
Fixes: https://github.com/aspnet/DataProtection/issues/73

On IIS where there is no user profile, the code tries to always read the 32bit registry view irrespective of the bitness of the worker process. So in case of 64 bit app pools the registry key is null so it falls back to in memory ephemeral repository. On 32 bit app pool it can find an appropriate registry key, but the keyRepositoryDescriptor is not populated resulting in a null reference exception.

Current behavior:
X86 throws
X64 falls back to in memory ephemeral

With fix:
Both X86 and X64 will use DPAPI.

@davidfowl @ChengTian @glennc @anurse @muratg


Eilon:
:shipit:

PR #78, opened 2015-04-07
Add serviceable attribute to projects.

NTaylorMullen:


src/Microsoft.AspNet.DataProtection.Shared/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Reflection;
+
+[assembly: AssemblyMetadata("Serviceable", "True")]
```
Eilon: I think this project isn't a real project, right? Build-time only.

NTaylorMullen: :+1:

NTaylorMullen:
Updated.

Eilon:
:shipit:

PR #79, opened 2015-04-17
Use $HOME as a possible storage location

### LINKED ISSUE #76: FileSystemXmlRepository depends on Windows environment variables when running on CoreCLR
In [FileSystemXmlRepository.cs, Line 106](https://github.com/aspnet/DataProtection/blob/dev/src/Microsoft.AspNet.DataProtection/Repositories/FileSystemXmlRepository.cs#L106) the code assumes that either LOCALAPPDATA or USERPROFILE exists.  If neither does, an ArugmentException will be thrown when ```Path.Combine``` is invoked, with ```null``` as an argument (since ```Environment.GetEnvironmentVarible``` returns null if the variable is not preset).

One could imagine changing the code to look like this:

```csharp
            // On core CLR, we need to fall back to environment variables.
            var localAppDataPath = Environment.GetEnvironmentVariable("LOCALAPPDATA");
            var userProfilePath = Environment.GetEnvironmentVariable("USERPROFILE");

            // On non-windows, these environment varibles are normally not present
            if (localAppDataPath == null && userProfilePath == null)
            {
                return null;
            }

            var folderPath = localAppDataPath ?? Path.Combine(userProfilePath, "AppData", "Local");
            DirectoryInfo retVal = GetKeyStorageDirectoryFromBaseAppDataPath(folderPath);
            try
            {
                retVal.Create(); // throws if we don't have access, e.g., user profile not loaded
                return retVal;
            }
            catch
            {
                return null;
            }
```

But it's not clear to me that's the right thing long term.

ellismg:
If LOCALAPPDATA and USERPROFILE are both null (as is the case on
Linux/OSX) use "$HOME/.aspnet" as the root folder for data protection
keys

Fixes #76

### DIFF
--- a/src/Microsoft.AspNet.DataProtection/Repositories/FileSystemXmlRepository.cs
+++ b/src/Microsoft.AspNet.DataProtection/Repositories/FileSystemXmlRepository.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Xml.Linq;
@@ -61,9 +62,11 @@ namespace Microsoft.AspNet.DataProtection.Repositories
         /// </summary>
         protected IServiceProvider Services { get; }
 
+        private const string DataProtectionKeysFolderName = "DataProtection-Keys";
+
         private static DirectoryInfo GetKeyStorageDirectoryFromBaseAppDataPath(string basePath)
         {
-            return new DirectoryInfo(Path.Combine(basePath, "ASP.NET", "DataProtection-Keys"));
+            return new DirectoryInfo(Path.Combine(basePath, "ASP.NET", DataProtectionKeysFolderName));
         }
 
         public virtual IReadOnlyCollection<XElement> GetAllElements()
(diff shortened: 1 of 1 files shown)


msotclas:
Hi __@ellismg__, I'm your friendly neighborhood Microsoft Open Technologies, Inc. Pull Request Bot (You can call me MSOTBOT). Thanks for your contribution!
    <p>
        It looks like you're a Microsoft contributor (Matt Ellis). If you're full-time, we DON'T require a Contribution License Agreement. If you are a vendor, please DO sign the electronic Contribution License Agreement. It will take 2 minutes and there's no faxing! https://cla2.msopentech.com.
    </p>

TTYL, MSOTBOT;

ellismg:
@davidfowl @GrabYourPitchforks Thoughts on this approach?  I didn't do platform sniffing because it was not clear I could always get an IRuntimeEnvironment from the service provider and it didn't seem worth the effort to do the old pinvoke to libc::uname trick.

GrabYourPitchforks:
Looks reasonable to me. :)

ellismg:
Thanks @GrabYourPitchforks.  @davidfowl The Travis build is broken but it looks like it's been that way for a while.  What can I do to get this ready to ship?  I would like to get the fix on myget so I can stop doing my hacky workaround of setting LOCALAPPDATA.

ellismg:
@Eilon Levi thought you might be a good guy to tell me if this is ready to ship or not.

src/Microsoft.AspNet.DataProtection/Repositories/FileSystemXmlRepository.cs
```diff
@@ -61,9 +62,11 @@ public FileSystemXmlRepository([NotNull] DirectoryInfo directory, IServiceProvid
         /// </summary>
         protected IServiceProvider Services { get; }
 
+        const string DataProtectionKeysFolderName = "DataProtection-Keys";
```
Eilon: Add `private` modifier to this.

Eilon:
:shipit: Looks legitimate, just have one small comment, then merge.

ellismg:
Thanks, @Eilon!

Issue #91, opened 2015-08-15
Skip building Microsoft.AspNet.DataProtection.SystemWeb on mono


davidfowl:
Mono doesn't support enough to build this properly so we should update the sake script to skip this one on mono

PR #95, opened 2015-10-07
Fix build on Linux

cesarbs:
Summary of the changes
 - Skip building Microsoft.AspNet.DataProtection.SystemWeb
 - Fix/disable unit tests

#91

### DIFF
--- a/makefile.shade
+++ b/makefile.shade
@@ -1,5 +1,6 @@
 use assembly='WindowsBase, Version=4.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35'
 use namespace='System.IO.Packaging'
+use import="Environment"
 
 var VERSION='0.1'
 var FULL_VERSION='0.1'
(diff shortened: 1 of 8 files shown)


cesarbs:
@muratg @davidfowl @Tratcher @victorhurdugaci

README.md
```diff
@@ -1,8 +1,8 @@
 DataProtection
```
Tratcher: Revert?

makefile.shade
```diff
@@ -1,45 +1,108 @@
-use assembly='WindowsBase, Version=4.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35'
```
Tratcher: ??? We should look at your git config autocrlf settings

test/Microsoft.AspNet.DataProtection.Extensions.Test/DataProtectionProviderTests.cs
```diff
@@ -2,8 +2,12 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections;
+using System.Collections.Generic;
 using System.IO;
```
Tratcher: You don't need most of these usings do you?

makefile.shade
```diff
@@ -1,45 +1,108 @@
...
+                        @"src\Microsoft.AspNet.DataProtection.SystemWeb\web.config.transform",
+                        @"content\net451\web.config.transform");
+                }
+            }
+        }
+    }
+
+functions
+    @{
+        PackagePart CreatePartFromFile(
```
davidfowl: We should be able to get rid of this and use the new feature that allows packing any file into the nupkg. See https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNet.Server.Kestrel/project.json#L62

test/Microsoft.AspNet.DataProtection.Test/Managed/ManagedAuthenticatedEncryptorTests.cs
```diff
@@ -77,7 +79,8 @@ public void Encrypt_Decrypt_Tampering_Fails()
             });
         }
 
-        [Fact]
+        [ConditionalFact]
+        [ConditionalRunTestOnlyOnWindows]
```
Tratcher: Comment? What's Windows specific about this one?

test/Microsoft.AspNet.DataProtection.Test/Managed/ManagedAuthenticatedEncryptorTests.cs
```diff
@@ -94,7 +97,7 @@ public void Encrypt_KnownKey()
             byte[] retVal = encryptor.Encrypt(
                 plaintext: plaintext,
                 additionalAuthenticatedData: aad);
-
+
```
Tratcher: nit: revert

test/Microsoft.AspNet.DataProtection.Test/Managed/ManagedAuthenticatedEncryptorTests.cs
cesarbs: To be honest, I'm not sure. I tried getting the same results on Linux but had no success on that. Then I noticed that there are tests with the same name in other classes and those were already marked to run on Windows only.

makefile.shade
cesarbs: Fixed.

Tratcher:
:shipit:

PR #96, opened 2015-10-08
Replace NotNullAttribute with thrown exceptions

pranavkm:


cesarbs:
What is the reason for this change?

pranavkm:
We're removing support for NotNull attribute since DNX won't have support to make it up using precompilation in v1.

cesarbs:
:shipit:

Issue #94, opened 2015-09-24
Logging: Add event ids to all publication sites 


anpete:
Any calls to ILogger.Log should specify a non-zero event id that is unique within the current logging category.

PR #105, opened 2015-11-13
Add event ids to all publication sites

pakrym:
Issue: https://github.com/aspnet/DataProtection/issues/94
I renamed logging extension methods to be more EF like.

@anpete @muratg @victorhurdugaci @moozzyk


src/Microsoft.AspNet.DataProtection/AuthenticatedEncryption/CngCbcAuthenticatedEncryptionOptions.cs
```diff
@@ -154,7 +154,7 @@ private BCryptAlgorithmHandle GetSymmetricBlockCipherAlgorithmHandle(ILogger log
 
             if (logger.IsVerboseLevelEnabled())
             {
-                logger.LogVerboseF($"Opening CNG algorithm '{EncryptionAlgorithm}' from provider '{EncryptionAlgorithmProvider}' with chaining mode CBC.");
+                logger.LogVerbose(DataProtectionEventId.CngCbcAuthenticatedEncryptionOptions, $"Opening CNG algorithm '{EncryptionAlgorithm}' from provider '{EncryptionAlgorithmProvider}' with chaining mode CBC.");
```
anpete: Event ids should typically be unique within a single logging "category". Suggest you create specific ids for Hmac and SymmetricBlockCipher code paths.

anpete:
Per comment, ensure unique event ids per logger category. Also, consider taking ILogger<T> as ctor dependency parameter, rather than passing logger instance to individual methods.

victorhurdugaci:
Our team uses a different style for logging. Take a look at these previous PRs that add logging:

- https://github.com/aspnet/Mvc/commit/da731fc641f56529328d3cb74d6092310a40627e
- https://github.com/aspnet/StaticFiles/pull/76
- https://github.com/aspnet/Hosting/pull/447

src/Microsoft.AspNet.DataProtection/DataProtectionEventId.cs
```diff
@@ -0,0 +1,27 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNet.DataProtection
+{
+    public enum DataProtectionEventId
+    {
+        KeyServices = 1,
```
victorhurdugaci: See the PRs that I mentioned for our logging style. It's more clear

src/Microsoft.AspNet.DataProtection/DataProtectionEventId.cs
```diff
@@ -0,0 +1,27 @@
...
+        XmlKeyManager,
+        KeyRingProvider,
+        CertificateXmlEncryptor,
+        DpapiNGXmlEncryptor,
+        DpapiNGXmlDecryptor,
```
victorhurdugaci: Another con for enums: if you remove an item, all those that follow will get a different value

src/Microsoft.AspNet.DataProtection/LoggingExtensions.cs
```diff
@@ -69,44 +70,72 @@ private static bool IsLogLevelEnabledCore(ILogger logger, LogLevel level)
...
+        {
+            logger.LogWarning((int)eventId, message);
+        }
+
+        public static void LogWarning(this ILogger logger, DataProtectionEventId eventId, Exception error, FormattableString message)
+        {
+            logger.LogWarning((int)eventId, new FormattedLogValues(message.Format, message.GetArguments()), error);
+        }
+
+        public static void LogWarning(this ILogger logger, DataProtectionEventId eventId, Exception error, string message)
```
victorhurdugaci: By using simple constants there's no need to cast

pakrym:
@victorhurdugaci 

In MVC we don't log eventids 

In StaticFiles we have LoggerExtensions class with logging messages set as delegate fields and eventId just hardcoded in them.

In Hosting we have LoggerExtensions class with log call's and level check in method bodies and event ids in seperate LoggerEventIds class with const fields. And we don't log eventids in some places (https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNet.Server.Testing/Common/RetryHelper.cs)

So, what is our style for logging?

victorhurdugaci:
MVC has event ids: https://github.com/aspnet/Mvc/blob/da731fc641f56529328d3cb74d6092310a40627e/src/Microsoft.AspNet.Mvc.Core/Logging/HttpStatusCodeResultLoggerExtensions.cs#L17

Use the delegates version when you have arguments to pass. Use the other when you don't. See this bug: https://github.com/aspnet/Logging/issues/272

The pattern is:
1. Have the messages as extension methods for `ILogger`
2. Have (ideally) all event ids in one place
3. Make sure it's not possible to create the same message with a different id (that's why we hide the message and error level in the method)
4. Encapsulate the `if (log level enabled) { log(..) }` logic

pakrym:
@victorhurdugaci  at the same time https://github.com/aspnet/Mvc/blob/da731fc641f56529328d3cb74d6092310a40627e/src/Microsoft.AspNet.Mvc.Core/Controllers/FilterActionInvoker.cs#L300 how do we decide if message needs eventId?

pakrym:
We changed them later, I see

victorhurdugaci:
I don't think MVC did a pass through all their messages to add event ids. They should all have ids.

Event ids should be ideally unique. That's why having them in one place makes things easy. Also, they shouldn't change once set, that's why an enum might not be ideal.

Eilon:
Yeah the pattern seen in https://github.com/aspnet/Mvc/blob/da731fc641f56529328d3cb74d6092310a40627e/src/Microsoft.AspNet.Mvc.Core/Logging/HttpStatusCodeResultLoggerExtensions.cs is the correct pattern. We have plenty of work items logged for us to update places where we don't yet use that new pattern. The new pattern follows all the logging best practices, and is also highly performant.

Issue #102, opened 2015-11-11
DataProtection no longer uses the correct default when hosted in the IIS/Kestrel combination


blowdart:
When I start an IIS hosted app Data Protection self configured as Ephemeral keys, that is it holds a key chain in memory and throws it away once the app stops. This is a huge regression from Helios.

You can see this by turning on verbose logging
 
```
warning : [Microsoft.Framework.DependencyInjection.DataProtectionServices] Neither user profile nor HKLM registry available. Using an ephemeral key repository. Protected data will be unavailable when application exits.
warning : [Microsoft.AspNet.DataProtection.Repositories.EphemeralXmlRepository] Using an in-memory repository. Keys will not be persisted to storage.
```

The expected behavior is store keys in the registry, protected using machine level DPAPI. 

If the same app is run via web.cmd you get what I would expect because you’re running under a user profile. IIS Express also has the correct behavior, again because a user profile is available

```
info    : [Microsoft.Framework.DependencyInjection.DataProtectionServices] Userprofile is available. Using 'C:\Users\bdorrans\AppData\Local\ASP.NET\DataProtection-Keys' as key repository and Windows DPAPI to encrypt keys at rest.
```

The configuration code lies in DataProtectionServices.cs

blowdart:
For now you can fix it by changing the app pool to load the user profile. This will then act like IIS Express, and use the file system.

pakrym:
Cause of issue is that `SOFTWARE\Microsoft\ASP.NET\4.0.30319.0\AutoGenKeys` does not exist.
Cause of that can be:
1. We are running wrong bitness of dnx vs w3wp
2. App domain doesn't have CLR set for it so IIS doesn't create a key for us.

pakrym:
Script that seems to solve the issue: https://github.com/aspnet/DataProtection/pull/106

muratg:
@blowdart What should we do with this bug?  Is it only providing documentation and scripts, plus trying to get IIS implement what we want? If so, can we close this one?

Putting it in 1.0.1 for now as there doesn't seem to be an action in rtm.

blowdart:
Yup, for now it's scripts. The request to fix it is in with the IIS folks, so we can bump it, and keep the pressure on.

muratg:
Closing this bug as there's no code change that can fix this on our stack. My understanding is this is purely an IIS fix that may or may not happen

PR #106, opened 2015-11-14
Add script to add AutoGenKeys section and UCL for it

pakrym:
Based on https://support.microsoft.com/en-us/kb/2915218

Usage `.\Provision-AutoGenKeys.ps1 DefaultAppPool`
@muratg @blowdart @Eilon

### DIFF
--- a/Provision-AutoGenKeys.ps1
+++ b/Provision-AutoGenKeys.ps1
@@ -1,0 +1,82 @@
+param (
+    [Parameter(Mandatory = $True)]
+    [string] $appPoolName
+  )
+
+# Provisions the HKLM registry so that the specified user account can persist auto-generated machine keys.
+function Provision-AutoGenKeys {
+  [CmdletBinding()]
+  param (
+    [ValidateSet("2.0", "4.0")]
+    [Parameter(Mandatory = $True)]
+    [string] $frameworkVersion,
+    [ValidateSet("32", "64")]
+    [Parameter(Mandatory = $True)]
+    [string] $architecture,
+    [Parameter(Mandatory = $True)]
+    [string] $sid
+  )
+  process {
+    # We require administrative permissions to continue.
+    if (-Not (new-object System.Security.Principal.WindowsPrincipal([System.Security.Principal.WindowsIdentity]::GetCurrent())).IsInRole([System.Security.Principal.WindowsBuiltInRole]::Administrator)) {
+        Write-Error "This cmdlet requires Administrator permissions."
+        return
+    }
+    # Open HKLM with an appropriate view into the registry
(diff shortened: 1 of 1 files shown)


blowdart:
Feel free to merge this, or talk to @HaoK if you're feeling nervous. Hao this doesn't need to become part of a package or anything, but it does need to live in the repo.

davidfowl:
@blowdart Is this a short term solution? This relies on full .NET and ASP.NET 4.x being enabled right?

pakrym:
@davidfowl  No, it should create keys and set acl for them.

blowdart:
It's a short term fix for #102 - but it doesn't rely on anything. We just need to ensure people run this manually, which sucks. For RC2 we need a much better solution where the registry keys are created automatically.

davidfowl:
How does this work https://github.com/aspnet/DataProtection/pull/106/files#diff-5458ff3cb68d9ae62ebd612f39237b5bR38? Doesn't it rely on the ASP.NET base key being created?

pakrym:
True,  will add a check for upper levels

blowdart:
Ah yea, just creating it blindly should be fine. Should still have that name, for .net 4 & 5 cookie sharing

blowdart:
Also does WOW64 get in the way here? Need to test for both 32 and 64 bit

GrabYourPitchforks:
@davidfowl WAS creates that registry key automatically when creating an app pool which is marked as "CLR v4.x".

Eilon:
Looks a'ight to me, as best I can tell.

@anurse to review the PS code because it's PS code and he knows PS code and I don't know PS code :smile:

pakrym:
I updated script to check for asp.net folder

anurse:
Yep, it looks like PS code to me :). :shipit: from that perspective alone

davidfowl:
@GrabYourPitchforks the concern is that .NET Framework isn't always available. I just want to make sure that if we're relying on it, it's a short term solution.

pakrym:
@davidfowl the only thing is, this script requires .net 4 and PS 3

muratg:
Yes, this has to be a short term solution.  @pakrym is assigned to #102 for the correct fix in RC2.

GrabYourPitchforks:
That registry key was only ever intended to be used in the Helios hosting case. You'd obviously need to move off of it now that Helios has been eliminated as a hosting option.

poke:
I just ran into the issue that I was trying to run this on a machine that didn’t have PowerShell 3, so I ported the script into an actual executable. So in case anyone else needs this: https://github.com/poke/aspnet-ProvisionAutoGenKeys

PR #109, opened 2015-11-17
ConfigureDP => AddDP

HaoK:
cc @Tratcher This differs a bit from all the other repos, but I just left things mostly alone and turned Configure into an overload of Add like everyone else.


Tratcher:
:shipit:

PR #110, opened 2015-11-17
Add event ids to all log calls

pakrym:
https://github.com/aspnet/DataProtection/issues/94

@victorhurdugaci @moozzyk @anpete


src/Microsoft.AspNet.DataProtection/LoggingExtensions.cs
```diff
@@ -13,6 +14,264 @@ namespace Microsoft.Extensions.Logging
...
+
+        private static Action<ILogger, Guid, Exception> _triedToProcessRevocationOfKeyButNoSuchKeyWasFound;
+
+        private static Action<ILogger, Guid, Exception> _foundKey;
+
+        private static Action<ILogger, DateTimeOffset, Exception> _foundRevocationOfAllKeysCreatedPriorTo;
+
+        private static Action<ILogger, Guid, Exception> _foundRevocationOfKey;
+
+        private static Action<ILogger, XElement, Exception> _ExceptionWhileProcessingRevocationElement;
```
Eilon: Oh no, field starts with capital `E` :smile:

src/Microsoft.AspNet.DataProtection/LoggingExtensions.cs
```diff
@@ -13,6 +14,264 @@ namespace Microsoft.Extensions.Logging
     /// </summary>
     internal static class LoggingExtensions
     {
+        private static Action<ILogger, Guid, DateTimeOffset, Exception> _usingFallbackKeyWithExpirationAsDefaultKey;
```
victorhurdugaci: Unrelated, we should use a generator (T4?) to generate these classes. It's a pain to maintain them

src/Microsoft.AspNet.DataProtection/LoggingExtensions.cs
```diff
@@ -69,44 +328,350 @@ private static bool IsLogLevelEnabledCore(ILogger logger, LogLevel level)
...
 
-        public static void LogDebugF(this ILogger logger, Exception error, FormattableString message)
+        public static void WritingDataToFile(this ILogger logger, string finalFilename)
         {
-            logger.LogDebug(new FormattedLogValues(message.Format, message.GetArguments()), error);
+            _writingDataToFile(logger, finalFilename, null);
         }
 
-        public static void LogError(this ILogger logger, Exception error, string message)
+        public static void ReadingDataFromRegistryKeyValue(this ILogger logger, Win32.RegistryKey regKey, string valueName)
```
victorhurdugaci: Use namespace?

victorhurdugaci:
Looks good :shipit: assuming the CI passes

pakrym:
:up: :date:
Requires https://github.com/aspnet/Logging/pull/296

src/Microsoft.AspNet.DataProtection/LoggingExtensions.cs
```diff
@@ -69,44 +421,294 @@ private static bool IsLogLevelEnabledCore(ILogger logger, LogLevel level)
             return (logger != null && logger.IsEnabled(level));
         }
 
-        public static void LogDebugF(this ILogger logger, FormattableString message)
+        public static void UsingFallbackKeyWithExpirationAsDefaultKey(this ILogger logger, Guid keyId, DateTimeOffset expirationDate)
+        {
+            _usingFallbackKeyWithExpirationAsDefaultKey(logger, keyId, expirationDate, null);
```
Eilon: Named param for the `null` at the end of these?

Eilon:
Just small comments, then :shipit:

src/Microsoft.AspNet.DataProtection/LoggingExtensions.cs
pakrym: Nope, its Action<?> so param name would be argN

Eilon: Oh right, I keep forgetting that.

moozzyk: good ol' inline comment `/* parameterX */ null`

GrabYourPitchforks:
You're going to have to rerun perf tests against this library. The code was originally written in such a way that gathering the data to be written to the log was deferred until it was known for certain that there was a logger listening. For instance, the expensive `JoinPurposesForLog` operation was never called unless the was a logger present and it had the *debug* level enabled. This method is now called for all calls to Protect and Unprotect, even if no logger is present. (There are other instances of expensive operations, but this is the most obvious.)

/cc @blowdart

pakrym:
@ GrabYourPitchforks I left checks in that places

GrabYourPitchforks:
@pakrym great :)

PR #111, opened 2015-11-20
Fixed possibly wrong registry path

dkraemer:
Your version of this script opens/creates registry path "HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\4.0.30319.0\AutoGenKeys\$sid"
I think the correct path should be "HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\ASP.NET\4.0.30319.0\AutoGenKeys\$sid"


Eilon:
@pakrym @blowdart

pakrym:
Thank you!

PR #112, opened 2015-12-07
Reacting to verbose rename

JunTaoLuo:
Reacting to https://github.com/aspnet/Logging/pull/314.
Tested locally.

### DIFF
--- a/src/Microsoft.AspNet.DataProtection/LoggingExtensions.cs
+++ b/src/Microsoft.AspNet.DataProtection/LoggingExtensions.cs
@@ -147,43 +147,43 @@ namespace Microsoft.Extensions.Logging
                 formatString: "Policy resolution states that a new key should be added to the key ring, but automatic generation of keys is disabled. Using fallback key {KeyId:B} with expiration {ExpirationDate:u} as default key.");
             _usingKeyAsDefaultKey = LoggerMessage.Define<Guid>(
                 eventId: 2,
-                logLevel: LogLevel.Verbose,
+                logLevel: LogLevel.Debug,
                 formatString: "Using key {KeyId:B} as the default key.");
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNet.DataProtection/LoggingExtensions.cs
```diff
@@ -421,13 +421,13 @@ public static bool IsInformationLevelEnabled(this ILogger logger)
...
         /// <summary>
-        /// Returns a value stating whether the 'verbose' log level is enabled.
+        /// Returns a value stating whether the 'trace' log level is enabled.
         /// Returns false if the logger instance is null.
         /// </summary>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static bool IsVerboseLevelEnabled(this ILogger logger)
+        public static bool IsTraceLevelEnabled(this ILogger logger)
         {
-            return IsLogLevelEnabledCore(logger, LogLevel.Verbose);
```
kichalla: Why is this not `Debug`? or an intentional behavior change?

kichalla: Based on your other PR (logging) comment, this makes sense to me now...but the above `LoggingExtensions.cs` changes rename `Verbose` to `Debug`?

JunTaoLuo: This is an api change. There is already an `IsDebugLevelEnabled` function. The only thing that needs to change here is `Verbose` to `Trace` so every level has a corresponding helper function.

JunTaoLuo:
@kichalla I was wondering if you still had questions for this one?

kichalla:
:shipit:

PR #119, opened 2016-02-18
Enabled xml doc generation

ajaybhargavb:
@NTaylorMullen

### DIFF
--- a/src/Microsoft.AspNetCore.DataProtection.Abstractions/IDataProtectionProvider.cs
+++ b/src/Microsoft.AspNetCore.DataProtection.Abstractions/IDataProtectionProvider.cs
@@ -11,7 +11,7 @@ namespace Microsoft.AspNetCore.DataProtection
         /// <summary>
         /// Creates an <see cref="IDataProtector"/> given a purpose.
         /// </summary>
-        /// <param name="purposes">
+        /// <param name="purpose">
         /// The purpose to be assigned to the newly-created <see cref="IDataProtector"/>.
         /// </param>
         /// <returns>An IDataProtector tied to the provided purpose.</returns>
--- a/src/Microsoft.AspNetCore.DataProtection/DataProtectionConfiguration.cs
+++ b/src/Microsoft.AspNetCore.DataProtection/DataProtectionConfiguration.cs
@@ -17,6 +17,7 @@ using System.Security.Cryptography.X509Certificates;
 
 namespace Microsoft.AspNetCore.DataProtection
 {
+#if !DOTNET5_4
     /// <summary>
     /// Provides access to configuration for the data protection system, which allows the
     /// developer to configure default cryptographic algorithms, key storage locations,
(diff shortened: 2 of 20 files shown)


src/Microsoft.AspNetCore.DataProtection/DataProtectionConfiguration.cs
```diff
@@ -26,7 +26,7 @@ namespace Microsoft.AspNetCore.DataProtection
     /// <para>
     /// If the developer changes the at-rest key protection mechanism, it is intended that
     /// he also change the key storage location, and vice versa. For instance, a call to
-    /// <see cref="ProtectKeysWithCertificate(string)"/> should generally be accompanied by
+    /// <c>ProtectKeysWithCertificate(string)</c> should generally be accompanied by
```
NTaylorMullen: See should work fine here. Is this the Roslyn bug?

ajaybhargavb: This was causing failures because the method `ProtectKeysWithCertificate` is defined inside an `ifdef` and is reference outside. @blowdart, any suggestions on how to change this comment without referencing this method?

cc' @dougbu

blowdart: ifdef the comments too so it's right on whatever platform? It needs to be there in supported platforms

ajaybhargavb: Well, we can't ifdef a specific part of the comment. The only way is to ifdef the entire `<summary>` and we'll have two summary with just one line difference.

src/Microsoft.AspNetCore.DataProtection.Abstractions/IDataProtectionProvider.cs
```diff
@@ -11,7 +11,7 @@ public interface IDataProtectionProvider
         /// <summary>
         /// Creates an <see cref="IDataProtector"/> given a purpose.
         /// </summary>
-        /// <param name="purposes">
+        /// <param name="purpose">
```
Eilon: Did you make this change on porpoise :dolphin: ?

src/Microsoft.AspNetCore.DataProtection/DataProtectionConfiguration.cs
dougbu: @blowdart when @ajaybhargavb asked me about this comment, my hope was we'd have something to follow `For instance` that was not platform-specific. Are you dashing my hopes (again)?

blowdart: Yup, consider yourself dashed.

src/Microsoft.AspNetCore.DataProtection.Abstractions/IDataProtectionProvider.cs
ajaybhargavb: Yes. :fishing_pole_and_fish:

ajaybhargavb:
:up: :date:

NTaylorMullen:
:ship: :eyes: :tea:

PR #123, opened 2016-02-25
[Fixes #121, Fixes #120] Make data protection follow the options model

### LINKED ISSUE #120: Make Microsoft.AspNetCore.DataProtection.DataProtectionProvider class static
https://github.com/aspnet/DataProtection/blob/dev/src/Microsoft.AspNetCore.DataProtection.Extensions/DataProtectionProvider.cs

https://docs.asp.net/en/latest/security/data-protection/configuration/non-di-scenarios.html

This class is meant to be used in Non-DI scenarios as an entry point for creating a IDataProtectionProvider instance that can be used acrross the app to protect and unprotect data. As such, making it static will avoid any confusion of mixing it with DI enabled scenarios. The advantages are:

* Clearly separates DI scenarios from non DI scenarios. (You can't register a static class in a DI Container)
* Makes the class cleaner (The current implementation creates a whole data protection stack inside the constructor, and wraps the default implementation without any perceived benefit.
* Is arguably less discoverable, it's easier to 'dot' into a static class and see a method that you can use to construct an instance of the thing that you want, than having to use 'new' on the type.
* By doing all the setup in a constructor, we are cornering ourselves in the future if we wan't to do more complicated things inside the setup, like for example, caching existing DataProtectionProviders.

As a reference, here is the implementation as a static class
```c#
using System;
using System.IO;
using Microsoft.Extensions.DependencyInjection;

namespace Microsoft.AspNetCore.DataProtection
{
    public static class DataProtectionProvider
    {
        public static IDataProtectionProvider Create(DirectoryInfo keyDirectory)
        {
            return Create(keyDirectory, setupAction: null);
        }

        public static IDataProtectionProvider Create(
            DirectoryInfo keyDirectory,
            Action<DataProtectionConfiguration> setupAction)
        {
            if (keyDirectory == null)
            {
                throw new ArgumentNullException(nameof(keyDirectory));
            }

            var serviceCollection = new ServiceCollection();
            var builder = serviceCollection.AddDataProtection();
            builder.PersistKeysToFileSystem(keyDirectory);

            if(setupAction != null)
            {
                setupAction(builder);
            }

            return serviceCollection.BuildServiceProvider().GetRequiredService<IDataProtectionProvider>();
        }
    }
}
```

### LINKED ISSUE #121: Make DataProtection follow the Options model
This repo requires some cleanup to adapt it to the pattern we follow on other parts of the framework

javiercn:


javiercn:
Things to pay attention to:
* Comments (There's been a lot of copy paste, so mistakes can happen).
* Method names
* Null checks for arguments
* Overload with configuration action calls parameterless overload.
* Parameterless overload does the work and brings in direct depenendencies
  (Like routing on MVC, for most repositories it means making sure AddOptions is called).

HaoK:
Don't introduce IDataProtectionBuilder here either

Eilon:
@HaoK what's the concern here? This seems to follow the interface-builder pattern we have elsewhere, no? Is the concern with this specific change, or with changing it at all?

HaoK:
The interface/builder pattern was only in MVC, and not anywhere else (not here/EF/or identity).  The interface pattern seems more suited for MVC which has tons of assemblies layering on methods, and just an extra interface everywhere else...

@divega and I pushed back on identity for that reason, and I was just echoing that comment here when I saw a builder interface added.

davidfowl:
It would be good to look at some samples of this being called before going further.

javiercn:
@davidfowl 

```
services
  .AddDataProtection(options =>{...})
  .PersistXXX()
  .UseFileXXX()
```

javiercn:
@blowdart

blowdart:
Data Protection isn't just for MVC. If the builder is an MVC only thing then it needs to go. 

How does it cope with options that are mutually exclusive? Or if you try to set the same thing twice?

Eilon:
Builders aren't an MVC thing. I'm setting up a meeting for Thursday to discuss this pattern in general.

javiercn:
@Eilon @rynowak I believe I've incorporated all the changes we discussed on the meeting

javiercn:
@Eilon ping? 😀

src/Microsoft.AspNetCore.DataProtection.Extensions/DataProtectionProvider.cs
```diff
@@ -8,21 +8,19 @@
 namespace Microsoft.AspNetCore.DataProtection
 {
     /// <summary>
-    /// A simple implementation of an <see cref="IDataProtectionProvider"/> where keys are stored
-    /// at a particular location on the file system.
+    /// An entry point class for creating an <see cref="IDataProtectionProvider"/> where keys are stored
+    /// at a particular location on the file system on non dependency injection enabled scenarios.
```
Eilon: I'd put the `on non dependency injection enabled scenarios` part in a `<remarks>` section. Also, the phrasing is a bit nerdy. Perhaps something like `<remarks>Use this method when not using dependency injection to provide the service to the application.</remarks>`?

src/Microsoft.AspNetCore.DataProtection/DataProtectionBuilderExtensions.cs
```diff
@@ -17,196 +17,195 @@
...
+    public static class DataProtectionBuilderExtensions
     {
         /// <summary>
-        /// Creates a new configuration object linked to a <see cref="IServiceCollection"/>.
+        /// Sets the unique name of this application within the data protection system.
         /// </summary>
-        public DataProtectionConfiguration(IServiceCollection services)
+        /// <param name="builder">The <see cref="IDataProtectionBuilder"/>.</param>
+        /// <param name="applicationName">The application name.</param>
+        /// <returns>The 'this' instance.</returns>
```
Eilon: Yuck. Copy what we do elsewhere. I hope we don't do this anywhere else.

src/Microsoft.AspNetCore.DataProtection/DataProtectionBuilderExtensions.cs
```diff
@@ -424,61 +474,73 @@ public DataProtectionConfiguration UseCustomCryptographicAlgorithms(CngGcmAuthen
...
                 }
             }
         }
 
-        private void Use(ServiceDescriptor descriptor)
+        private static void Use(IServiceCollection services, ServiceDescriptor descriptor)
         {
-            RemoveAllServicesOfType(descriptor.ServiceType);
-            Services.Add(descriptor);
+            RemoveAllServicesOfType(services, descriptor.ServiceType);
```
Eilon: In the meeting didn't we say something about this method? Like why does it exist? Is it needed?

src/Microsoft.AspNetCore.DataProtection/DataProtectionBuilderExtensions.cs
Eilon: Looks like elsewhere in this repo we do the more correct "returns a reference to this instance after the operation completed" or whatever - so let's use that everywhere here.

Eilon:
Sorry for the delay. Looks like it's 99% :shipit: , just a few things.

src/Microsoft.AspNetCore.DataProtection.Extensions/DataProtectionProvider.cs
rynowak: > An entry point class for creating an <see cref="IDataProtectionProvider"/> where keys are stored at a particular location on the file system on non dependency injection enabled scenarios.

Suggest:

> Contains factory methods for creating an <see cref="IDataProtectionProvider"/> where keys are stored at a particular location in the file system for non-dependency-injection scenarios.

src/Microsoft.AspNetCore.DataProtection.Extensions/DataProtectionProvider.cs
```diff
@@ -31,9 +29,11 @@ public DataProtectionProvider(DirectoryInfo keyDirectory)
...
         /// <param name="keyDirectory">The <see cref="DirectoryInfo"/> in which keys should be stored. This may
         /// represent a directory on a local disk or a UNC share.</param>
-        /// <param name="configure">An optional callback which provides further configuration of the data protection
-        /// system. See <see cref="DataProtectionConfiguration"/> for more information.</param>
-        public DataProtectionProvider(DirectoryInfo keyDirectory, Action<DataProtectionConfiguration> configure)
+        /// <param name="setupAction">An optional callback which provides further configuration of the data protection
+        /// system. See <see cref="IDataProtectionBuilder"/> for more information.</param>
+        public static IDataProtectionProvider Create(
+            DirectoryInfo keyDirectory,
+            Action<IDataProtectionBuilder> setupAction)
```
rynowak: Didn't you do a pass to make these callbacks non-optional in a bunch of places? :trollface:

src/Microsoft.AspNetCore.DataProtection/AuthenticatedEncryption/ConfigurationModel/AuthenticatedEncryptorDescriptor.cs
```diff
@@ -8,22 +8,22 @@ namespace Microsoft.AspNetCore.DataProtection.AuthenticatedEncryption.Configurat
...
     /// A descriptor which can create an authenticated encryption system based upon the
-    /// configuration provided by an <see cref="AuthenticatedEncryptionOptions"/> object.
+    /// configuration provided by an <see cref="AuthenticatedEncryptionSettings"/> object.
     /// </summary>
     public sealed class AuthenticatedEncryptorDescriptor : IAuthenticatedEncryptorDescriptor
     {
         private readonly IServiceProvider _services;
 
-        public AuthenticatedEncryptorDescriptor(AuthenticatedEncryptionOptions options, ISecret masterKey)
+        public AuthenticatedEncryptorDescriptor(AuthenticatedEncryptionSettings options, ISecret masterKey)
```
rynowak: options -> settings

src/Microsoft.AspNetCore.DataProtection/AuthenticatedEncryption/ConfigurationModel/CngCbcAuthenticatedEncryptorConfiguration.cs
```diff
@@ -13,23 +13,23 @@ public sealed class CngCbcAuthenticatedEncryptorConfiguration : IAuthenticatedEn
     {
         private readonly IServiceProvider _services;
 
-        public CngCbcAuthenticatedEncryptorConfiguration(CngCbcAuthenticatedEncryptionOptions options)
+        public CngCbcAuthenticatedEncryptorConfiguration(CngCbcAuthenticatedEncryptionSettings options)
```
rynowak: options -> settings

src/Microsoft.AspNetCore.DataProtection/AuthenticatedEncryption/ConfigurationModel/CngCbcAuthenticatedEncryptorDescriptor.cs
```diff
@@ -9,22 +9,22 @@ namespace Microsoft.AspNetCore.DataProtection.AuthenticatedEncryption.Configurat
...
     /// A descriptor which can create an authenticated encryption system based upon the
-    /// configuration provided by an <see cref="CngCbcAuthenticatedEncryptionOptions"/> object.
+    /// configuration provided by an <see cref="CngCbcAuthenticatedEncryptionSettings"/> object.
     /// </summary>
     public sealed class CngCbcAuthenticatedEncryptorDescriptor : IAuthenticatedEncryptorDescriptor
     {
         private readonly ILogger _log;
 
-        public CngCbcAuthenticatedEncryptorDescriptor(CngCbcAuthenticatedEncryptionOptions options, ISecret masterKey)
+        public CngCbcAuthenticatedEncryptorDescriptor(CngCbcAuthenticatedEncryptionSettings options, ISecret masterKey)
```
rynowak: options -> settings

src/Microsoft.AspNetCore.DataProtection/AuthenticatedEncryption/ConfigurationModel/CngGcmAuthenticatedEncryptorConfiguration.cs
```diff
@@ -13,23 +13,23 @@ public sealed class CngGcmAuthenticatedEncryptorConfiguration : IAuthenticatedEn
     {
         private readonly IServiceProvider _services;
 
-        public CngGcmAuthenticatedEncryptorConfiguration(CngGcmAuthenticatedEncryptionOptions options)
+        public CngGcmAuthenticatedEncryptorConfiguration(CngGcmAuthenticatedEncryptionSettings options)
```
rynowak: Noticing a pattern here

src/Microsoft.AspNetCore.DataProtection/AuthenticatedEncryption/ConfigurationModel/CngGcmAuthenticatedEncryptorDescriptor.cs
```diff
@@ -9,22 +9,22 @@ namespace Microsoft.AspNetCore.DataProtection.AuthenticatedEncryption.Configurat
...
     /// A descriptor which can create an authenticated encryption system based upon the
-    /// configuration provided by an <see cref="CngGcmAuthenticatedEncryptionOptions"/> object.
+    /// configuration provided by an <see cref="CngGcmAuthenticatedEncryptionSettings"/> object.
     /// </summary>
     public sealed class CngGcmAuthenticatedEncryptorDescriptor : IAuthenticatedEncryptorDescriptor
     {
         private readonly ILogger _log;
 
-        public CngGcmAuthenticatedEncryptorDescriptor(CngGcmAuthenticatedEncryptionOptions options, ISecret masterKey)
+        public CngGcmAuthenticatedEncryptorDescriptor(CngGcmAuthenticatedEncryptionSettings options, ISecret masterKey)
```
rynowak: :sleeping:

src/Microsoft.AspNetCore.DataProtection/AuthenticatedEncryption/ConfigurationModel/ManagedAuthenticatedEncryptorConfiguration.cs
```diff
@@ -14,23 +14,23 @@ public sealed class ManagedAuthenticatedEncryptorConfiguration : IAuthenticatedE
     {
         private readonly IServiceProvider _services;
 
-        public ManagedAuthenticatedEncryptorConfiguration(ManagedAuthenticatedEncryptionOptions options)
+        public ManagedAuthenticatedEncryptorConfiguration(ManagedAuthenticatedEncryptionSettings options)
```
rynowak: OK, officially not commenting on this one.

src/Microsoft.AspNetCore.DataProtection/AuthenticatedEncryption/ConfigurationModel/ManagedAuthenticatedEncryptorDescriptorDeserializer.cs
```diff
@@ -42,7 +42,7 @@ public IAuthenticatedEncryptorDescriptor ImportFromXml(XElement element)
             //   <masterKey>...</masterKey>
             // </descriptor>
 
-            var options = new ManagedAuthenticatedEncryptionOptions();
+            var options = new ManagedAuthenticatedEncryptionSettings();
```
rynowak: this one is different though

src/Microsoft.AspNetCore.DataProtection/DataProtectionBuilderExtensions.cs
```diff
@@ -17,196 +17,195 @@
...
 namespace Microsoft.AspNetCore.DataProtection
 {
-#if !NETSTANDARD1_3
     /// <summary>
-    /// Provides access to configuration for the data protection system, which allows the
-    /// developer to configure default cryptographic algorithms, key storage locations,
-    /// and the mechanism by which keys are protected at rest.
+    /// Extensions for configuring data protection using an <see cref="IDataProtectionBuilder"/>.
     /// </summary>
-    /// <remarks>
```
rynowak: Why delete all this stuff?

src/Microsoft.AspNetCore.DataProtection/DataProtectionBuilderExtensions.cs
rynowak: This would usually say something like `A reference to the <see cref="IDataProtectionBuilder" /> after this operation has completed.`

src/Microsoft.AspNetCore.DataProtection/DataProtectionBuilderExtensions.cs
```diff
@@ -424,61 +474,73 @@ public DataProtectionConfiguration UseCustomCryptographicAlgorithms(CngGcmAuthen
...
+                throw new ArgumentNullException(nameof(builder));
+            }
+
+            Use(builder.Services, DataProtectionServiceDescriptors.IDataProtectionProvider_Ephemeral());
+            return builder;
         }
 
         /*
          * UTILITY ISERVICECOLLECTION METHODS
          */
```
rynowak: This makes me a sad panda :panda_face:

src/Microsoft.AspNetCore.DataProtection/DataProtectionBuilderExtensions.cs
```diff
@@ -17,196 +17,195 @@
...
         /// <summary>
-        /// Configures miscellaneous global options.
+        /// Configures the key management options for the data protection system.
         /// </summary>
-        /// <param name="setupAction">A callback that configures the global options.</param>
+        /// <param name="builder">The <see cref="IDataProtectionBuilder"/>.</param>
+        /// <param name="setupAction">An <see cref="Action{KeyManagementOptions}"/> to configure the provided<see cref="KeyManagementOptions"/>.</param>
         /// <returns>The 'this' instance.</returns>
-        public DataProtectionConfiguration ConfigureGlobalOptions(Action<DataProtectionOptions> setupAction)
+        public static IDataProtectionBuilder AddKeyManagement(this IDataProtectionBuilder builder, Action<KeyManagementOptions> setupAction)
```
rynowak: This should be `AddKeyManagementOptions`

src/Microsoft.AspNetCore.DataProtection/DataProtectionBuilderExtensions.cs
```diff
@@ -17,196 +17,195 @@
...
-            if (services == null)
+            if (builder == null)
             {
-                throw new ArgumentNullException(nameof(services));
+                throw new ArgumentNullException(nameof(builder));
             }
 
-            Services = services;
-        }
+            builder.Services.Configure<DataProtectionOptions>(options =>
```
rynowak: There should also be an `AddDataProtectionOptions` somewhere in here

src/Microsoft.AspNetCore.DataProtection/DataProtectionServiceDescriptors.cs
```diff
@@ -58,13 +58,13 @@ public static ServiceDescriptor ConfigureOptions_DefaultKeyLifetime(int numDays)
         /// </summary>
         public static ServiceDescriptor IAuthenticatedEncryptorConfiguration_Default()
         {
-            return IAuthenticatedEncryptorConfiguration_FromOptions(new AuthenticatedEncryptionOptions());
+            return IAuthenticatedEncryptorConfiguration_FromOptions(new AuthenticatedEncryptionSettings());
```
rynowak: From settings?

src/Microsoft.AspNetCore.DataProtection/DataProtectionBuilderExtensions.cs
Eilon: It's not deleted, it moved verbatim to the builder interface file.

rynowak: :+1: - I don't read things, I just go with my gut.

Eilon: Don't worry, it caught me out too the first time I saw the code :smile:

src/Microsoft.AspNetCore.DataProtection/DataProtectionBuilderExtensions.cs
javiercn: Yes, but it doesn't affect public API and I'm trying not to change behavior. We know it works and we can fix it later if we want, given that it's private and just 10 lines of code, I don't feel it's worth fixing at this point.

src/Microsoft.AspNetCore.DataProtection/DataProtectionBuilderExtensions.cs
javiercn: It was called ConfigureGlobalOptions and we decided to get rid of it.

src/Microsoft.AspNetCore.DataProtection.Extensions/DataProtectionProvider.cs
javiercn: Thisi s addressed

src/Microsoft.AspNetCore.DataProtection/DataProtectionBuilderExtensions.cs
```diff
@@ -0,0 +1,546 @@
...
+            return builder;
+        }
+
+        /// <summary>
+        /// Configures the key management options for the data protection system.
+        /// </summary>
+        /// <param name="builder">The <see cref="IDataProtectionBuilder"/>.</param>
+        /// <param name="setupAction">An <see cref="Action{KeyManagementOptions}"/> to configure the provided<see cref="KeyManagementOptions"/>.</param>
+        /// <returns>A reference to the <see cref="IDataProtectionBuilder" /> after this operation has completed.</returns>
+        public static IDataProtectionBuilder AddKeyManagement(this IDataProtectionBuilder builder, Action<KeyManagementOptions> setupAction)
```
rynowak: `AddKeyManagementOptions`

rynowak:
:shipit:

PR #127, opened 2016-03-11
Target dnx451

bricelam:
This reverts 9750924973bd1f99841cde7c801c4d0888e195fb for DNX


muratg:
Do we still need these, since we got the newer CLI today?

@pranavkm @anurse

bricelam:
Recalling. After chatting with @Eilon, we decided that adding the facades to user's `project.json` is an acceptable workaround.

PR #131, opened 2016-03-17
Build only if src directory exists.

cesarbs:
Required for CI testing - we delete src/ folders to speed up the process.

cc @victorhurdugaci @pakrym


makefile.shade
```diff
@@ -11,7 +11,7 @@ var Configuration_Local = '${E("Configuration")}'
 var ROOT_Local = '${Directory.GetCurrentDirectory()}'
 var BUILD_DIR_Local = '${Path.Combine(ROOT_Local, "build")}'
 
-#build-compile target='compile' if='IsLinux'
+#build-compile target='compile' if='IsLinux && Directory.Exists("src")'
```
pakrym: Why do we have this target at all?

victorhurdugaci: Because some projects can't compile on Linux

Issue #134, opened 2016-03-24
Refactor DI support in DataProtection


ajaybhargavb:

muratg:
@ajaybhargavb How close is this for check-in?

@Eilon Are we taking this in RC2?  Or postpone to 1.0.0?

ajaybhargavb:
I'll be sending an update to the PR today. But it will not include and test and documentation changes.

Eilon:
I'd like to get this for RC2 because there are public API changes in it, and I spoke to @ajaybhargavb about it and we had a short email thread too.

muratg:
Looks like the PR is a little stale at the moment. Do we have any blockers on this?

ajaybhargavb:
Spoke to @Eilon and @rynowak this morning. This is being moved to RTM.

ajaybhargavb:
The latest code and samples are in https://github.com/aspnet/DataProtection/tree/ajbaaska/new-dataprotection branch.

GuardRex:
@ajaybhargavb You're working on the DI bits, right? I ask because I was having a small problem with DI into the `XmlDecryptor`. https://github.com/aspnet/DataProtection/issues/154

The problem was that DI for `XmlDecryptor` wasn't working. I wanted to do this ...
```
public CustomXmlDecryptor(IOptions<CustomDataProtectionOptions> dataProtectionOptions)
{
    _key = dataProtectionOptions.Value.Key;
}
```
... but that wasn't working. I simply wasn't getting the value. It works using the ctor injection pattern for `XmlEncryptor`, just not for `XmlDecryptor`. 

With a little help from @tillig, I was able to make it work on the decryptor side with a little hack, I ended up having to do this ...
```
public CustomXmlDecryptor(IServiceProvider services)
{
    var dataProtectionOptions = 
        services.GetRequiredService<IOptions<CustomDataProtectionOptions>>();
    _key = dataProtectionOptions.Value.Key;
}
```
Is fixing this something that falls under this issue, or was this problem already addressed do you know?

rynowak:
@GuardRex - please make a separate issue for this. Having looked at the code, this isn't something that's been supported so far so it's a feature request

GuardRex:
@rynowak Ok, thanks. Will do.

GuardRex:
[...]

PR #136, opened 2016-03-28
[Design] Refactor DataProtection

ajaybhargavb:
Issue - #134 

@rynowak @loudej @lodejard 

Things that still need cleanup:
1. DataProtectionServiceDescriptors
2. CertificateXmlEncryptor
3. EncryptedXmlDescriptor


src/Microsoft.AspNetCore.DataProtection/KeyManagement/XmlKeyManager.cs
```diff
@@ -454,5 +451,65 @@ void IInternalXmlKeyManager.RevokeSingleKey(Guid keyId, DateTimeOffset revocatio
...
+                    sink.Store(keyId, element);
+                }
+            }
+        }
+
+        /// <summary>
+        /// A simplified default implementation of <see cref="IActivator"/> that understands
+        /// how to call ctors which take <see cref="ILoggerFactory"/>.
+        /// </summary>
+        private sealed class SimpleActivator : IActivator
```
ajaybhargavb: This needs to be removed. Needs some discussion on how to approach this scenario.

src/Microsoft.AspNetCore.DataProtection/ActivatorExtensions.cs
```diff
@@ -1,91 +0,0 @@
...
-using Microsoft.AspNetCore.Cryptography;
-using Microsoft.AspNetCore.DataProtection.Internal;
-using Microsoft.Extensions.DependencyInjection;
-
-namespace Microsoft.AspNetCore.DataProtection
-{
-    /// <summary>
-    /// Extension methods for working with <see cref="IActivator"/>.
-    /// </summary>
-    internal static class ActivatorExtensions
```
rynowak: I wasn't around for the resolution of this - are we confident that we can/should remove this feature?

src/Microsoft.AspNetCore.DataProtection/AuthenticatedEncryption/ConfigurationModel/AuthenticatedEncryptorDescriptor.cs
```diff
@@ -33,7 +29,7 @@ public AuthenticatedEncryptorDescriptor(AuthenticatedEncryptionSettings settings
 
             Settings = settings;
             MasterKey = masterKey;
-            _services = services;
+            _loggerFactory = loggerFactory;
```
rynowak: Where's it used?

src/Microsoft.AspNetCore.DataProtection/DataProtectionProvider.cs
```diff
@@ -0,0 +1,90 @@
...
+            var keyRingProvider = new KeyRingProvider(keyManager, keyManagementOptions, loggerFactory, defaultKeyResolver, cacheableKeyRingProvider: null);
+
+            IDataProtectionProvider dataProtectionProvider = new KeyRingBasedDataProtectionProvider(keyRingProvider, loggerFactory);
+            if (!string.IsNullOrEmpty(dataProtectionOptions.ApplicationDiscriminator))
+            {
+                dataProtectionProvider = dataProtectionProvider.CreateProtector(dataProtectionOptions.ApplicationDiscriminator);
+            }
+
+            return dataProtectionProvider;
+        }
```
rynowak: This all looks like decent cleanup

src/Microsoft.AspNetCore.DataProtection/EphemeralDataProtectionProvider.cs
```diff
@@ -48,10 +48,10 @@ public EphemeralDataProtectionProvider(IServiceProvider services)
                 keyringProvider = new EphemeralKeyRing<ManagedAuthenticatedEncryptionSettings>();
             }
 
-            var logger = services.GetLogger<EphemeralDataProtectionProvider>();
+            var logger = loggerFactory?.CreateLogger<EphemeralDataProtectionProvider>();
             logger?.UsingEphemeralDataProtectionProvider();
```
rynowak: This is the kind of thing that I was talking about when we discussed in person earlier - this logging statement doesn't include any detail about the guts of this class, it just states what provider we're using. That logging could come from anywhere

src/Microsoft.AspNetCore.DataProtection/KeyManagement/KeyEscrowServiceProviderExtensions.cs
```diff
@@ -1,42 +0,0 @@
...
-        /// Gets an aggregate <see cref="IKeyEscrowSink"/> from the underlying <see cref="IServiceProvider"/>.
-        /// This method may return null if no sinks are registered.
-        /// </summary>
-        public static IKeyEscrowSink GetKeyEscrowSink(this IServiceProvider services)
-        {
-            var escrowSinks = services?.GetService<IEnumerable<IKeyEscrowSink>>()?.ToList();
-            return (escrowSinks != null && escrowSinks.Count > 0) ? new AggregateKeyEscrowSink(escrowSinks) : null;
-        }
-
-        private sealed class AggregateKeyEscrowSink : IKeyEscrowSink
```
rynowak: Where did this go? unused?

src/Microsoft.AspNetCore.DataProtection/KeyManagement/XmlKeyManager.cs
```diff
@@ -71,40 +85,19 @@ public sealed class XmlKeyManager : IKeyManager, IInternalXmlKeyManager
...
-
-            // If the repository is missing, then we get both the encryptor and the repository from the fallback.
-            // If the fallback is missing, the final call to GetRequiredService below will throw.
-            if (KeyRepository == null)
+            if (escrowSinks != null && escrowSinks.Any())
             {
-                var defaultKeyServices = services.GetService<IDefaultKeyServices>();
-                KeyEncryptor = defaultKeyServices?.GetKeyEncryptor(); // optional
-                KeyRepository = defaultKeyServices?.GetKeyRepository() ?? services.GetRequiredService<IXmlRepository>();
+                _keyEscrowSink = new AggregateKeyEscrowSink(escrowSinks);
```
rynowak: Ah there it is

src/Microsoft.AspNetCore.DataProtection/XmlEncryption/XmlEncryptionExtensions.cs
```diff
@@ -48,7 +48,7 @@ public static XElement DecryptElement(this XElement element, IActivator activato
                 var clonedElementWhichRequiresDecryption = new XElement(elementWhichRequiresDecryption);
                 var innerDoc = new XDocument(clonedElementWhichRequiresDecryption);
                 string decryptorTypeName = (string)clonedElementWhichRequiresDecryption.Attribute(XmlConstants.DecryptorTypeAttributeName);
-                var decryptorInstance = activator.CreateInstance<IXmlDecryptor>(decryptorTypeName);
+                var decryptorInstance = activator.CreateInstance(typeof(IXmlDecryptor), decryptorTypeName) as IXmlDecryptor;
```
rynowak: do you expect this to ever fail the cast? It should be a c-style cast.

src/Microsoft.AspNetCore.DataProtection/ActivatorExtensions.cs
ajaybhargavb: This is not really being removed but moved to `XmlKeyManager` which is the only place this is needed.

src/Microsoft.AspNetCore.DataProtection/XmlEncryption/XmlEncryptionExtensions.cs
ajaybhargavb: Will change

src/Microsoft.AspNetCore.DataProtection/EphemeralDataProtectionProvider.cs
ajaybhargavb: Yes,  this call can be moved outside this class but it still needs the `ILoggerFactory` which is passed in here https://github.com/aspnet/DataProtection/pull/136/files#diff-46b998be1bdad1ab2c77fb0e3b161b7aR54. I am still checking if we can switch to using some kind of context class.

src/Microsoft.AspNetCore.DataProtection/DataProtectionProvider.cs
```diff
@@ -0,0 +1,91 @@
...
+using Microsoft.AspNetCore.DataProtection.Cng;
+using Microsoft.AspNetCore.DataProtection.KeyManagement;
+using Microsoft.AspNetCore.DataProtection.Repositories;
+using Microsoft.AspNetCore.DataProtection.XmlEncryption;
+using Microsoft.Extensions.Logging;
+using Microsoft.Win32;
+
+namespace Microsoft.AspNetCore.DataProtection
+{
+    public abstract class DataProtectionProvider : IDataProtectionProvider
```
PinpointTownes: Won't this type conflict with the one used by the extensions package?

ajaybhargavb: I guess one of these will probably have to be renamed.

ajaybhargavb:
:up: :date:

src/Microsoft.AspNetCore.DataProtection/DataProtectionProvider.cs
PinpointTownes: If you plan to include this PR in RC2, maybe it would be better to address that now?

ajaybhargavb: Yes. But this PR is still in design. I am planning to do all the documentation/test/renames in a later update.

src/Microsoft.AspNetCore.DataProtection/ActivatorExtensions.cs
```diff
@@ -37,55 +37,8 @@ public static T CreateInstance<T>(this IActivator activator, string implementati
         public static IActivator GetActivator(this IServiceProvider serviceProvider)
```
rynowak: This is strange as an extension method. Services are one of those things that you should know if it's null or not.

src/Microsoft.AspNetCore.DataProtection/ActivatorExtensions.cs
```diff
@@ -37,55 +37,8 @@ public static T CreateInstance<T>(this IActivator activator, string implementati
...
-                {
-                    return ctorWhichTakesServiceProvider.Invoke(new[] { _services });
-                }
-
-                // Finally, prefer .ctor() as an ultimate fallback.
-                // This will throw if the ctor cannot be called.
-                return Activator.CreateInstance(implementationType);
-            }
+                ? (serviceProvider.GetService<IActivator>() ?? new SimpleActivator(serviceProvider.GetService<ILoggerFactory>()))
+                : new SimpleActivator(loggerFactory: null);
```
rynowak: why null logger factory? I think the first thing we need to do is stop the logger factory from being so viral in this code.

src/Microsoft.AspNetCore.DataProtection/ActivatorExtensions.cs
```diff
@@ -37,55 +37,8 @@ public static T CreateInstance<T>(this IActivator activator, string implementati
...
-                if (ctorWhichTakesServiceProvider != null)
-                {
-                    return ctorWhichTakesServiceProvider.Invoke(new[] { _services });
-                }
-
-                // Finally, prefer .ctor() as an ultimate fallback.
-                // This will throw if the ctor cannot be called.
-                return Activator.CreateInstance(implementationType);
-            }
+                ? (serviceProvider.GetService<IActivator>() ?? new SimpleActivator(serviceProvider.GetService<ILoggerFactory>()))
```
rynowak: In what case would we have no `IActivator` registered but `ILoggerFactory` registered? Is the activator a replaceable abstraction? (ask yourself if that makes sense).

src/Microsoft.AspNetCore.DataProtection/ActivatorExtensions.cs
```diff
@@ -37,55 +37,8 @@ public static T CreateInstance<T>(this IActivator activator, string implementati
         public static IActivator GetActivator(this IServiceProvider serviceProvider)
         {
             return (serviceProvider != null)
```
rynowak: This method would really benefit from using things like `if` and `else if`

src/Microsoft.AspNetCore.DataProtection/AuthenticatedEncryption/AuthenticatedEncryptionSettings.cs
```diff
@@ -55,10 +62,10 @@ public void Validate()
...
          */
 
-        internal IAuthenticatedEncryptor CreateAuthenticatedEncryptorInstance(ISecret secret, IServiceProvider services = null)
+        internal IAuthenticatedEncryptor CreateAuthenticatedEncryptorInstance(ISecret secret)
         {
             return CreateImplementationOptions()
-                .ToConfiguration(services)
+                .ToConfiguration()
                 .CreateDescriptorFromSecret(secret)
                 .CreateEncryptorInstance();
```
rynowak: So, this PR is updated now, what's the answer?

src/Microsoft.AspNetCore.DataProtection/ActivatorExtensions.cs
ajaybhargavb: This call is for unit testing

ajaybhargavb:
Closing this since the latest work is on a different branch

Issue #130, opened 2016-03-17
Consider adding a parameterless DataProtectionProvider.Create overload


PinpointTownes:
Currently, `DataProtectionProvider.Create` always overrides the default key repository by a file system repository pointing to the `keyDirectory` parameter.

That would be great if we had a parameterless overload that let the data protection block decide itself what's the most appropriate place to store the crypto keys (file system, registry, in-memory) instead of overriding it.

https://github.com/aspnet/DataProtection/blob/dev/src/Microsoft.AspNetCore.DataProtection.Extensions/DataProtectionProvider.cs

Having such an overload would make the authentication ticket serializer compat' package much easier to use in non-vNext apps.

@Eilon any chance we could have that for RC2? I'd really like to use that in the OWIN/Katana version of ASOS (see https://github.com/aspnet-contrib/AspNet.Security.OpenIdConnect.Server/issues/188).

/cc @blowdart @javiercn

GrabYourPitchforks:
FYI, the reason this overload didn't exist in the first place is because part of the security of the system hinges on the ability to isolate applications from one another, and outside the context of the ASP.NET hosting environment there's no good way to get this information automatically from the ambient environment. Forcing the directory to be specified manually provides this isolation. If you want an overload that doesn't require specifying a directory, then you really should at least force the developer to provide an application name or similar to regain the isolation that was lost. In the source code this is referred to as the "discriminator".

The "using DataProtection in ASP.NET 4.x applications" sample works around this by using the IIS metabase path as a unique identifier so that the developer doesn't need to worry about specifying an isolation identifier. Unfortunately this trick doesn't work for self-hosted ASP.NET 4.x apps.

/cc @blowdart

blowdart:
Oh dear. I did not realise this.

@ajaybhargavb time to change your work I'm afraid.

@PinpointTownes you're not going to get what you want. Is it worth doing this at all?

PinpointTownes:
> @PinpointTownes you're not going to get what you want. Is it worth doing this at all?

I've been asked to provide interoperability between ASOS vNext and the OWIN/Katana version. For  that, I need to use the same ticket format and the same data protection stack.

Why not adding a `discriminator` parameter, as suggested by @GrabYourPitchforks?

[...]

PR #138, opened 2016-03-31
Added a few DataProtectionProvider.Create overloads

ajaybhargavb:
Issue #130 

@blowdart


src/Microsoft.AspNetCore.DataProtection.Extensions/DataProtectionProvider.cs
```diff
@@ -49,13 +53,14 @@ public static IDataProtectionProvider Create(DirectoryInfo keyDirectory)
             // build the service collection
             var serviceCollection = new ServiceCollection();
             var builder = serviceCollection.AddDataProtection();
-            builder.PersistKeysToFileSystem(keyDirectory);
 
-            if (setupAction != null)
+            if (keyDirectory != null)
```
blowdart: This works? And falls through to registry, or azure directory or whatever? I'm kinda shocked. 


However, I'd still have expected null or as keyDirectory would throw an exception. In this method you would expect the parameter. 

So, refactor the service collection setup et al after to a private method, then put the argumentnullexception back and have both methods call into the private one.

ajaybhargavb: This works. It falls back to `DataProtectionServices.GetDefaultServices()` which contains all the default logic. Will add the null check.

ajaybhargavb:
:up: :date:

src/Microsoft.AspNetCore.DataProtection.Extensions/DataProtectionProvider.cs
```diff
@@ -40,21 +49,25 @@ public static IDataProtectionProvider Create(DirectoryInfo keyDirectory)
...
+
+        private static IDataProtectionProvider CreateProvider(
+            DirectoryInfo keyDirectory,
+            Action<IDataProtectionBuilder> setupAction)
+        {
             // build the service collection
             var serviceCollection = new ServiceCollection();
             var builder = serviceCollection.AddDataProtection();
+
             builder.PersistKeysToFileSystem(keyDirectory);
```
PinpointTownes: Like @blowdart, I'm a bit surprised this method doesn't throw if the parameter is null. Anyway, I'd personally add a little `if` here, to ensure it's never called when `keyDirectory` is null.

PinpointTownes:
Thanks! :shipit:

src/Microsoft.AspNetCore.DataProtection.Extensions/DataProtectionProvider.cs
ajaybhargavb: @PinpointTownes, this is a private method. When `keyDirectory` is null it will be caught here https://github.com/aspnet/DataProtection/pull/138/files#diff-ea6431c57866be038e1a9a1d98b14916R48

PinpointTownes: Sure, but there's no point calling `builder.PersistKeysToFileSystem` with a `null` directory, which is what happens when using the parameterless `Create` overload.

ajaybhargavb: Ah. I see :+1:

PinpointTownes: Heeeeeh! `PersistKeysToFileSystem` throws an exception when using a null value: https://github.com/aspnet/DataProtection/blob/dev/src/Microsoft.AspNetCore.DataProtection/DataProtectionBuilderExtensions.cs#L180.

Does the new overload even works?

ajaybhargavb: It worked before my `Addressed feedback` commit. Totally missed this during the refactor.

PinpointTownes: Haha, that's @blowdart's fault... silly Barry :smile:

blowdart: Hey you agreed :p

blowdart:
Needs tests.

ajaybhargavb:
:up: :date:

ajaybhargavb:
Added two more overloads that take a `X509Certificate2` as a parameter.

test/Microsoft.AspNetCore.DataProtection.Extensions.Test/DataProtectionProviderTests.cs
```diff
@@ -38,6 +38,46 @@ public void System_UsesProvidedDirectory()
         [ConditionalFact]
         [ConditionalRunTestOnlyIfLocalAppDataAvailable]
         [ConditionalRunTestOnlyOnWindows]
+        public void System_UsesDefaultDirectory()
+        {
+            var keysPath = Path.Combine(Environment.ExpandEnvironmentVariables("%LOCALAPPDATA%"), "ASP.NET", "DataProtection-Keys");
```
davidfowl: This looks like a disaster waiting to happen 😄

ajaybhargavb: This was the best way to test this :smile: . Also it preserves the existing files in that folder. So its not that bad.

muratg:
Looks good to me...  @blowdart good to check-in?

src/Microsoft.AspNetCore.DataProtection.Extensions/DataProtectionProvider.cs
```diff
@@ -40,22 +50,83 @@ public static IDataProtectionProvider Create(DirectoryInfo keyDirectory)
...
+                throw new ArgumentNullException(nameof(setupAction));
+            }
+
+            return CreateProvider(keyDirectory, setupAction, certificate: null);
+        }
 
+#if !NETSTANDARD1_3 // [[ISSUE60]] Remove this #ifdef when Core CLR gets support for EncryptedXml
+        /// <summary>
+        /// Creates an <see cref="DataProtectionProvider"/> given a location at which to store keys
+        /// and a <see cref="X509Certificate2"/> for encryption.
```
PinpointTownes: I'm probably nit-picking, but I'd update this comment a bit to indicate that the certificate is used to encrypt the keys, not the protected data.

blowdart: Agree.

test/Microsoft.AspNetCore.DataProtection.Extensions.Test/DataProtectionProviderTests.cs
blowdart: Yea, no tests is a lot worse.

blowdart:
Good once the comment is updated

src/Microsoft.AspNetCore.DataProtection.Extensions/DataProtectionProvider.cs
```diff
@@ -15,6 +16,15 @@ namespace Microsoft.AspNetCore.DataProtection
     public static class DataProtectionProvider
     {
         /// <summary>
+        /// Creates a <see cref="DataProtectionProvider"/> that store keys on a location based on
```
rynowak: `on a location based` -> `in a location based`

src/Microsoft.AspNetCore.DataProtection.Extensions/DataProtectionProvider.cs
```diff
@@ -3,6 +3,7 @@
 
 using System;
 using System.IO;
+using System.Security.Cryptography.X509Certificates;
```
rynowak: ifdef this using as well

src/Microsoft.AspNetCore.DataProtection.Extensions/DataProtectionProvider.cs
```diff
@@ -40,22 +50,83 @@ public static IDataProtectionProvider Create(DirectoryInfo keyDirectory)
...
+        /// and a <see cref="X509Certificate2"/> for encryption.
+        /// </summary>
+        /// <param name="keyDirectory">The <see cref="DirectoryInfo"/> in which keys should be stored. This may
+        /// represent a directory on a local disk or a UNC share.</param>
+        /// <param name="certificate">The <see cref="X509Certificate2"/> to be used for encryption.</param>
+        public static IDataProtectionProvider Create(
+            DirectoryInfo keyDirectory,
+            X509Certificate2 certificate)
+        {
+            return Create(keyDirectory, setupAction: builder => { }, certificate: certificate);
```
rynowak: I don't see much value in chaining to the other overloads the public API. That just makes this harder to trace/examine, and makes these brittle to argument validation

rynowak: There's another call to the public API in `public static IDataProtectionProvider Create(DirectoryInfo keyDirectory)`

src/Microsoft.AspNetCore.DataProtection.Extensions/DataProtectionProvider.cs
```diff
@@ -40,22 +50,83 @@ public static IDataProtectionProvider Create(DirectoryInfo keyDirectory)
...
+            }
 
+            return CreateProvider(keyDirectory, setupAction, certificate);
+        }
+#endif
+
+        private static IDataProtectionProvider CreateProvider(
+            DirectoryInfo keyDirectory,
+            Action<IDataProtectionBuilder> setupAction,
+            X509Certificate2 certificate)
```
rynowak: Is it a problem to have this appear in the code for CoreCLR? (seems like it's not an issue, we just don't want the public API unless it's supported)

src/Microsoft.AspNetCore.DataProtection.Extensions/DataProtectionProvider.cs
rynowak: See comment here: https://github.com/aspnet/DataProtection/pull/138/files#r58582323

This might not matter

test/Microsoft.AspNetCore.DataProtection.Extensions.Test/DataProtectionProviderTests.cs
rynowak: This is pretty evil, but until the code is better factored we kinda need it. Do good tests for the other overloads?

test/Microsoft.AspNetCore.DataProtection.Extensions.Test/DataProtectionProviderTests.cs
```diff
@@ -38,6 +38,46 @@ public void System_UsesProvidedDirectory()
         [ConditionalFact]
         [ConditionalRunTestOnlyIfLocalAppDataAvailable]
         [ConditionalRunTestOnlyOnWindows]
+        public void System_UsesDefaultDirectory()
```
rynowak: Pick a better name

test/Microsoft.AspNetCore.DataProtection.Extensions.Test/DataProtectionProviderTests.cs
```diff
@@ -38,6 +38,46 @@ public void System_UsesProvidedDirectory()
         [ConditionalFact]
         [ConditionalRunTestOnlyIfLocalAppDataAvailable]
         [ConditionalRunTestOnlyOnWindows]
```
rynowak: Do we have a guarantee these won't run in parallel?

src/Microsoft.AspNetCore.DataProtection.Extensions/DataProtectionProvider.cs
ajaybhargavb: I don't think this is an issue.

test/Microsoft.AspNetCore.DataProtection.Extensions.Test/DataProtectionProviderTests.cs
```diff
@@ -38,6 +38,46 @@ public void System_UsesProvidedDirectory()
...
+                // Step 2: Instantiate the system and round-trip a payload
+                var protector = DataProtectionProvider.Create().CreateProtector("purpose");
+                Assert.Equal("payload", protector.Unprotect(protector.Protect("payload")));
+
+                // Step 3: Validate that there's now a single key in the directory and that it's protected using Windows DPAPI.
+                var newFileName = Assert.Single(Directory.GetFiles(keysPath));
+                var file = new FileInfo(newFileName);
+                Assert.StartsWith("key-", file.Name, StringComparison.OrdinalIgnoreCase);
+                var fileText = File.ReadAllText(file.FullName);
+                Assert.Contains("This key is encrypted with Windows DPAPI.", fileText, StringComparison.Ordinal);
```
rynowak: This this come from our code? or somewhere in teh windows?

ajaybhargavb: Our code. in `DpapiXmlEncryptor`

rynowak: OK, wanted to make sure this wouldn't break with a windows update or something

test/Microsoft.AspNetCore.DataProtection.Extensions.Test/DataProtectionProviderTests.cs
ajaybhargavb: I don't think that matters because there is no other test that can access this folder.

test/Microsoft.AspNetCore.DataProtection.Extensions.Test/DataProtectionProviderTests.cs
ajaybhargavb: We have tests for all overloads except the X509Certificate ones.

ajaybhargavb:
:up: :date:

PinpointTownes:
LGTM :smile: 

Question: should we have a `public static IDataProtectionProvider Create(string applicationName, Action<IDataProtectionBuilder> setupAction)` overload?

test/Microsoft.AspNetCore.DataProtection.Extensions.Test/DataProtectionProviderTests.cs
ajaybhargavb: Added a test for the X509Certificate overload.

src/Microsoft.AspNetCore.DataProtection.Extensions/DataProtectionProvider.cs
```diff
@@ -15,13 +16,37 @@ namespace Microsoft.AspNetCore.DataProtection
     public static class DataProtectionProvider
     {
         /// <summary>
+        /// Creates a <see cref="DataProtectionProvider"/> that store keys in a location based on
+        /// the platform and operating system.
+        /// </summary>
+        /// <param name="applicationName">An identifier that uniquely discriminates this application from all other
+        /// applications on the machine.</param>
+        public static IDataProtectionProvider Create(string applicationName)
```
rynowak: Will we need more overloads of this beyond the `Create(string)`. Will the one with options setup or a cert need to work?

test/Microsoft.AspNetCore.DataProtection.Extensions.Test/DataProtectionProviderTests.cs
```diff
@@ -63,6 +106,51 @@ public void System_UsesProvidedDirectory_WithConfigurationCallback()
             });
         }
 
+#if !NETSTANDARDAPP1_5 // [[ISSUE60]] Remove this #ifdef when Core CLR gets support for EncryptedXml
+        [ConditionalFact]
+        [ConditionalRunTestOnlyIfLocalAppDataAvailable]
+        [ConditionalRunTestOnlyOnWindows]
+        public void System_UsesProvidedDirectoryAndCertificate()
+        {
+            var filePath = Path.Combine(GetTestFilesPath(), "TestCert.pfx");
```
rynowak: Is this cert already part of the repo?

rynowak:
:shipit:

test/Microsoft.AspNetCore.DataProtection.Extensions.Test/DataProtectionProviderTests.cs
ajaybhargavb: Yes it is

src/Microsoft.AspNetCore.DataProtection.Extensions/DataProtectionProvider.cs
ajaybhargavb: Spoke to @blowdart. Added one with applicationName and the cert.

PR #141, opened 2016-04-25
Replace IApplicationEnvironment with IHostingEnvironment

pranavkm:


pranavkm:
cc @davidfowl

src/Microsoft.AspNetCore.DataProtection.Abstractions/DataProtectionExtensions.cs
```diff
@@ -120,15 +121,19 @@ public static IDataProtector CreateProtector(this IDataProtectionProvider provid
         [EditorBrowsable(EditorBrowsableState.Never)]
         public static string GetApplicationUniqueIdentifier(this IServiceProvider services)
         {
-            string discriminator = (services?.GetService(typeof(IApplicationDiscriminator)) as IApplicationDiscriminator)?.Discriminator;
-            if (discriminator == null)
+            string discriminator = null;
```
smbecker: Should this method be in Microsoft.AspNetCore.DataProtection so the .Abstractions project doesn't need a dependency on Hosting? That dependency seems unnecessary and undesirable for the Abstractions project, especially given that it is only used here.

src/Microsoft.AspNetCore.DataProtection.Abstractions/project.json
```diff
@@ -14,6 +14,8 @@
       "type": "build",
       "version": "1.0.0-*"
     },
+    "Microsoft.AspNetCore.Hosting.Abstractions": "1.0.0-*",
```
Eilon: Eek! Dependency on hosting? 😢 🐼

pranavkm: @smbecker had the same concern. I could move the method into the `DataProtection` to fix this. Alternatively, I could clean up how `GetApplicationUniqueIdentifier` works so that it doesn't have to know of `IApplicationEnvironment` \ `IHostingEnvironment`. I was about to file a RTM bug to clean up that bit.

pranavkm: Going with the "safer" change for RC2

Eilon:
:shipit:

PR #142, opened 2016-05-04
Fix DataProtection after build globbing changes will go in

pakrym:


makefile.shade
```diff
@@ -11,52 +11,55 @@ var Configuration_Local = '${E("Configuration")}'
...
             }
         }
-
-        DotnetPack(projectFile, BUILD_DIR_LOCAL, Configuration_Local);
+        DotnetPack(projectFile, BUILD_DIR_LOCAL, Configuration_Local, " --no-build");
     });
+    DotnetBuild(TEST_PROJECT_GLOB_LOCAL, Configuration_Local, BuildFramework);
 
-    foreach (var nupkg in Files.Include(Path.Combine(BUILD_DIR_LOCAL, "*/" + Configuration_Local + "/*.nupkg")))
+    if (srcProjects != null)
```
Eilon: This can't be null, can it?

pakrym: Yep, fixed.

PR #147, opened 2016-05-26
Don't break the CI.

cesarbs:
@pakrym


pakrym:
:shipit:

cesarbs:
I'll add a comment about this to the makefile. Easy to think it's some old thing that has no effect :smile:

PR #165, opened 2016-07-12
To de-dupe, we must remove

dougbu:
- part of aspnet/Common#93
- use `WebEncoders` from Common repo

Also let VS have its way w/ test `.xproj` files


anurse:
:shipit:

pranavkm:
:shipit:

dougbu:
4f30ddd

Issue #87, opened 2015-06-30
Allow KeyDerivation.Pbkdf2 to use a byte[] for the password


sqmgh:
Currently the function only allows passing a string to be hashed. It would be nice if you could pass a byte[] as well.

At the moment it seems kind of silly to be converting my byte[] to a string, so it can be converted back to a byte[]. Leaving a string floating around in GC land.

muratg:
This sounds like a good idea.

GrabYourPitchforks:
What's the scenario for this? Passwords are by definition textual input, thus their natural form is string rather than byte[].

sqmgh:
The requests stems from dealing with SecureString, though it wouldn't surprise me if there was another use case for it.

In the past since Rfc2898DeriveBytes worked with both string and byte[]. As it's much more simple to marshal to a byte[] and zero it afterwards than trying to pin a string in memory and clear that, it was decided to use a byte[].

So I suppose it's an XY problem. In my case having support for SecureString would solve the problem. However, due to there being very minimal amount of support for it, older methods allowing byte[], the code converting to a byte[] internally, and that *someone* probably has another use case for it. I requested byte[] support.

GrabYourPitchforks:
@sqmgh The SecureString API is essentially deprecated going forward and is slated for eventual removal from .NET Core. The general recommendation is that you switch to using simply String instead of SecureString throughout your application. Thus there's not really a huge benefit to providing a SecureString-based overload of this API.

See https://github.com/dotnet/apireviews/pull/8 for further discussion.

bwolkerstorfer:
@GrabYourPitchforks @sqmgh 

It would actually make sense to have the possibility to use a byte[] for migration/extension of legacy application environments

Background:
I am maintaining an application that used to call Rfc2898DeriveBytes with the result from ``System.Text.Encoding.Unicode.GetBytes(password)`` - the new implementation uses ``System.Text.Encoding.UTF8.GetBytes(password)``. The plan is to switch all existing services to .NET Core and that would force all users (quite a lot) to get a new password if we want to get rid of Rfc2898DeriveBytes.

blowdart:
As [Rfc2898DeriveBytes](https://github.com/dotnet/corefx/blob/d0dc5fc099946adc1035b34a8b1f6042eddb0c75/src/System.Security.Cryptography.Algorithms/src/System/Security/Cryptography/Rfc2898DeriveBytes.cs) still exists in core why can you not continue with what you're doing?

[...]

PR #168, opened 2016-07-20
KeyDerivation.Pbkdf2 use with byte[]

bwolkerstorfer:
change addresses #87 
- changed interfaces and providers PBKDF2 to support byte[] as input for password
- added tests for new logic


muratg:
cc @blowdart

blowdart:
We go to extraordinary lengths to pin memory, clear memory when we're done using it, etc. This PR deletes all of that work, and has a huge risk that passwords and byte arrays representing passwords hang around in memory long after they should have been cleared.

A further attack vector added with Encoding.UTF8.GetBytes(password). 

This PR is unsafe and must be rejected.

PinpointTownes:
@blowdart I have no particular interest for this new API signature, but I'd really appreciate if you could detail your reasoning (not everyone is a security expert :trollface:).

> We go to extraordinary lengths to pin memory, clear memory when we're done using it, etc. This PR deletes all of that work, and has a huge risk that passwords and byte arrays representing passwords hang around in memory long after they should have been cleared.

Since the password is initially flowed as a string, isn't that futile anyway? No matter what you do to reduce the lifetime of the intermediate byte arrays, the password is still held in clear as a string until it's garbage-collected. I'm not sure to understand how this PR makes this situation worse.

> A further attack vector added with Encoding.UTF8.GetBytes(password).

Aren't all the PBKDF2 providers (managed, Win7/Win8) already using it? AFAICT, only the data protection extensions use a UTF-8 encoding that throws on invalid byte sequences (if that's what you mean)
  - https://github.com/aspnet/DataProtection/blob/dev/src/Microsoft.AspNetCore.Cryptography.KeyDerivation/PBKDF2/ManagedPbkdf2Provider.cs#L71
  - https://github.com/aspnet/DataProtection/blob/dev/src/Microsoft.AspNetCore.Cryptography.KeyDerivation/PBKDF2/Win7Pbkdf2Provider.cs#L53
  - https://github.com/aspnet/DataProtection/blob/dev/src/Microsoft.AspNetCore.Cryptography.KeyDerivation/PBKDF2/Win8Pbkdf2Provider.cs#L86

blowdart:
We are pinning and clearing what we produced after we've done it. It's a best effort here, and surely you agree we ought to make an effort rather than ignoring it altogether.

PinpointTownes:
> It's a best effort here, and surely you agree we ought to make an effort rather than ignoring it altogether.

Sure (you could even use `SecureString` instead of a plain old `string`).

My question was: why do you think this PR is bad from a security perspective?

blowdart:
You know how I feel about securestring you troll :p

Removing the pinning and clearing of the array which we converted the string into means we now have a second copy of the the plaintext password floating about, which is obviously a bad idea.

PinpointTownes:
> You know how I feel about securestring you troll :p

Me, trolling? Nah (tho I've always wondered why you hated `SecureString` so much) :smile:

Actually, the DP stack heavily uses `Secret` - which seems to be an equivalent of `SecureString` that also relies on `CryptProtectMemory` - to "protect" critical things like DP master keys.

IMHO, user passwords are at least as sensitive as master keys, so it wouldn't surprise me if the PBKDF helpers accepted an `ISecret`/`SecureString`.

> Removing the pinning and clearing of the array which we converted the string into means we now have a second copy of the the plaintext password floating about, which is obviously a bad idea.

Sure, [this line](https://github.com/aspnet/DataProtection/pull/168/files#diff-a950609e3baa0d3f5d89e816dce57e0dR35) could be updated to use pinning and clear the intermediate bytes buffer before returning, but it's not clear to me why this makes the PR totally "unsafe".

Anyway, thanks for the details :smile:

PR #170, opened 2016-07-29
Some cleanup

ajaybhargavb:
@rynowak 

The first commit has `DataProtectionServiceDescriptors` cleanup
The second commit is just `var` cleanup


rynowak:
:shipit:

PR #178, opened 2016-09-28
Add KeyVault key protection support

pakrym:


blowdart:
Add the comment that every other protector has please, eg. https://github.com/aspnet/DataProtection/blob/dev/src/Microsoft.AspNetCore.DataProtection/XmlEncryption/DpapiNGXmlEncryptor.cs#L99

pakrym:
Closing until we decide on future of this.

Issue #181, opened 2016-10-03
Update project.json to use RTM version of StackExchange.Redis.StrongName


kichalla:
Recently this package was RTMed (http://www.nuget.org/packages/StackExchange.Redis.StrongName/): https://github.com/aspnet/DataProtection/blob/dev/src/Microsoft.AspNetCore.DataProtection.Redis/project.json#L17

@pakrym

PR #182, opened 2016-10-03
Update StackExchange.Redis.StrongName

pakrym:
Closes: https://github.com/aspnet/DataProtection/issues/181#event-810611216

@kichalla


kichalla:
Update the Redis sample to use netcore. After that :shipit:

PR #195, opened 2017-01-03
Implement IXmlRepository using IDistributedCache

cnblogs-dudu:
Like Microsoft.AspNetCore.Session, use IDistributedCache in Microsoft.Extensions.Caching.Distributed to store data. It is an approach for reusing the RedisCache implementation in Microsoft.Extensions.Caching.Redis.


blowdart:
This is an interesting idea, but we want to use concrete caches where we can guarantee behaviour or throw warnings. IDistributedMemoryCache & IMemoryCache are really aimed at in app caching, rather than to be generic for asp.net components.

Thanks though.

PR #203, opened 2017-02-22
Refactored DI support

ajaybhargavb:
Issue - #134 

- Refactored builder extensions and service collection extensions
- Refactored Settings/Configuration/Descriptor
- Removed ConfigurationCommon/AuthenticatedEncryptorConfigurationExtensions
- Added IAuthenticatedEncryptorFactory and implementations
- Refactored IKey to have Descriptor instead of CreateEncryptorInstance()
- Handled Repository/Encryptor special logic
- Added samples
- Updated tests

@rynowak @davidfowl @blowdart


GuardRex:
@ajaybhargavb Will these changes fix https://github.com/aspnet/DataProtection/issues/154?

TL;DR ... The prob described in that issue was that DI wasn't working for `IXmlDecryptor` ...
```c#
public class CustomXmlDecryptor : IXmlDecryptor
{
    private byte[] _key;

    public CustomXmlDecryptor()
    {
    }

    public CustomXmlDecryptor(IOptions<CustomDataProtectionOptions> dataProtectionOptions)
    {
        _key = dataProtectionOptions.Value.Key;
    }

    public XElement Decrypt(XElement encryptedElement)
    {
        // Code here uses _key to decrypt
    }
}
```
`_key` wasn't getting a value, so I've had to use ...
```c#
public CustomXmlDecryptor(IServiceProvider services)
{
    var dataProtectionOptions = 
        services.GetRequiredService<IOptions<CustomDataProtectionOptions>>();
    _key = dataProtectionOptions.Value.Key;
}
```
... to make it work.

ajaybhargavb:
@GuardRex, this change won't fix the issue that you referenced. It was never supported. `IXmlDecryptor` was not refactored as part of this change because it gets serialized and changing it would break existing applications.

test/Microsoft.AspNetCore.DataProtection.Test/RegistryPolicyResolverTests.cs
```diff
@@ -25,27 +26,36 @@ public class RegistryPolicyResolverTests
         [ConditionalRunTestOnlyIfHkcuRegistryAvailable]
         public void ResolvePolicy_NoEntries_ResultsInNoPolicies()
         {
-            IServiceCollection serviceCollection = new ServiceCollection();
-            RunTestWithRegValues(serviceCollection, new Dictionary<string, object>()
+            // Arrange
+            var registryEntries = new Dictionary<string, object>()
             {
                 ["unused"] = 42
```
Eilon: If it's unused, do we need to set any values in the registry at all?

src/Microsoft.AspNetCore.DataProtection/NullLoggerFactory.cs
```diff
@@ -0,0 +1,58 @@
...
+
+using System;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore.DataProtection
+{
+    /// <summary>
+    /// An <see cref="ILoggerFactory"/> that provides a logger that no ops.
+    /// </summary>
+    public class NullLoggerFactory : ILoggerFactory
```
Eilon: And... I looked at it some more, and is this used only in tests to begin with? I don't think I see it used at all in `src`, in which case at the very very least least it can move to the `test` folder as a utility library. But maybe it really should go into the Logging repo anywya.

ajaybhargavb: > is this used only in tests to begin with?

No. It is also used in `DataProtectionProviderFactory.GetDefaultLoggerFactory()`

> Should this move to https://github.com/aspnet/Logging/tree/dev/src/Microsoft.Extensions.Logging.Abstractions ?

Logging already has `NullLoggerFactory` in `Microsoft.Extensions.Logging.Testing`. https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging.Testing/NullLoggerFactory.cs. Should that be moved to `Microsoft.Extensions.Logging.Abstractions`?

test/Microsoft.AspNetCore.Cryptography.Internal.Test/Microsoft.AspNetCore.Cryptography.Internal.Test.csproj
```diff
@@ -21,4 +21,8 @@
     <PackageReference Include="xunit" Version="2.2.0-*" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />
```
ajaybhargavb: Yes, VS will just put this back.

src/Microsoft.AspNetCore.DataProtection/AuthenticatedEncryption/ManagedAuthenticatedEncryptorFactory.cs
```diff
@@ -0,0 +1,130 @@
...
+        }
+
+        private Func<SymmetricAlgorithm> GetSymmetricBlockCipherAlgorithmFactory(ManagedAuthenticatedEncryptorConfiguration configuration)
+        {
+            // basic argument checking
+            if (configuration.EncryptionAlgorithmType == null)
+            {
+                throw Error.Common_PropertyCannotBeNullOrEmpty(nameof(configuration.EncryptionAlgorithmType));
+            }
+            typeof(SymmetricAlgorithm).AssertIsAssignableFrom(configuration.EncryptionAlgorithmType);
```
ajaybhargavb: Talked offline

src/Microsoft.AspNetCore.DataProtection/NullLoggerFactory.cs
Eilon: Yeah I think we should consolidate all these `NullLogger***` types in one place. Can you log a bug in the Logging repo so we can clean this up, then consume it from here?

test/Microsoft.AspNetCore.Cryptography.Internal.Test/Microsoft.AspNetCore.Cryptography.Internal.Test.csproj
Eilon: Ah ok that's fine then.

src/Microsoft.AspNetCore.DataProtection/NullLoggerFactory.cs
ajaybhargavb: https://github.com/aspnet/Logging/issues/566

test/Microsoft.AspNetCore.DataProtection.Test/AuthenticatedEncryption/ConfigurationModel/CngGcmAuthenticatedEncryptorDescriptorDeserializerTests.cs
```diff
@@ -40,5 +43,19 @@ public void ImportFromXml_CreatesAppropriateDescriptor()
             byte[] roundTripPlaintext = test.Decrypt(new ArraySegment<byte>(ciphertext), new ArraySegment<byte>(aad));
             Assert.Equal(plaintext, roundTripPlaintext);
         }
+
+        private static IAuthenticatedEncryptor CreateEncryptorInstanceFromDescriptor(CngGcmAuthenticatedEncryptorDescriptor descriptor)
+        {
+            var key = new Key(
```
ajaybhargavb: Done

ajaybhargavb:
🆙 📅

src/Microsoft.AspNetCore.DataProtection/AuthenticatedEncryption/AuthenticatedEncryptorFactory.cs
```diff
@@ -1,108 +1,95 @@
...
+using Microsoft.AspNetCore.DataProtection.KeyManagement;
 using Microsoft.Extensions.Logging;
 
 namespace Microsoft.AspNetCore.DataProtection.AuthenticatedEncryption
 {
-    /// <summary>
-    /// Settings for configuring authenticated encryption algorithms.
-    /// </summary>
-    public sealed class AuthenticatedEncryptionSettings : IInternalAuthenticatedEncryptionSettings
+    public class AuthenticatedEncryptorFactory : IAuthenticatedEncryptorFactory
```
rynowak: Should this be sealed?

ajaybhargavb: Yes it should. Looks like I missed this one. Good catch 👍

ajaybhargavb:
Thanks @Eilon @rynowak @GrabYourPitchforks

rynowak:
🎉

Issue #220, opened 2017-04-05
Investigate: CreateEncryptorInstance runs multiple times


ajaybhargavb:
Comparing output of Azure sample,

1.0 release
```
dbug: Microsoft.AspNetCore.DataProtection.KeyManagement.XmlKeyManager[18]
      Found key {f0146984-4a7a-4a28-bfc0-cd3283e1b1a3}.
dbug: Microsoft.AspNetCore.DataProtection.KeyManagement.DefaultKeyResolver[13]
      Considering key {f0146984-4a7a-4a28-bfc0-cd3283e1b1a3} with expiration date 2017-07-04 17:24:35Z as default key.
dbug: Microsoft.AspNetCore.DataProtection.AuthenticatedEncryption.ConfigurationModel.CngCbcAuthenticatedEncryptorDescriptor[4]
      Opening CNG algorithm 'AES' from provider '(null)' with chaining mode CBC.
dbug: Microsoft.AspNetCore.DataProtection.AuthenticatedEncryption.ConfigurationModel.CngCbcAuthenticatedEncryptorDescriptor[3]
      Opening CNG algorithm 'SHA256' from provider '(null)' with HMAC.
dbug: Microsoft.AspNetCore.DataProtection.KeyManagement.KeyRingProvider[2]
      Using key {f0146984-4a7a-4a28-bfc0-cd3283e1b1a3} as the default key.
trce: Microsoft.AspNetCore.DataProtection.KeyManagement.KeyRingBasedDataProtector[31]
      Performing protect operation to key {f0146984-4a7a-4a28-bfc0-cd3283e1b1a3} with purposes ('sample-purpose').
CfDJ8IRpFPB6SihKv8DNMoPhsaN-ljs8dzZGN6D9MsYXlZesolbDKUgHLzkDaTQ6ppjkx6ke9jity-ZI_XtiXpl0w57yJFghk0Qlq_Wvf-XG-x4c7f8VnOUzDbgUmY4VOumZTA
```

Current:
```
dbug: Microsoft.AspNetCore.DataProtection.KeyManagement.XmlKeyManager[18]
      Found key {c5125edb-70b6-4e41-b4d8-8de8127812bf}.
dbug: Microsoft.AspNetCore.DataProtection.KeyManagement.DefaultKeyResolver[13]
      Considering key {c5125edb-70b6-4e41-b4d8-8de8127812bf} with expiration date 2017-07-04 19:30:00Z as default key.
[...]

PR #221, opened 2017-04-06
Creating only one IAuthenticatedEncryptor per IKey

ajaybhargavb:
Issue - #220 

Problem:
We should be creating only one `IAuthenticatedEncryptor` instance per key. After #134, we now have multiple instances being created.

Solution:
Brought back `CreateEncryptorInstance()` on `IKey`

Why was this needed?
Currently `IKey` and `IAuthenticatedEncryptor` creation are decoupled. Because of that there was no way to associate an encryptor instance with a key. Having `CreateEncryptorInstance()` on `IKey` lets us use the same encryptor instance across multiple calls.

@rynowak

### DIFF
--- a/src/Microsoft.AspNetCore.DataProtection/KeyManagement/IKey.cs
+++ b/src/Microsoft.AspNetCore.DataProtection/KeyManagement/IKey.cs
@@ -49,5 +49,12 @@ namespace Microsoft.AspNetCore.DataProtection.KeyManagement
         /// Gets the <see cref="IAuthenticatedEncryptorDescriptor"/> instance associated with this key.
         /// </summary>
         IAuthenticatedEncryptorDescriptor Descriptor { get; }
+
+        /// <summary>
+        /// Creates an <see cref="IAuthenticatedEncryptor"/> instance that can be used to encrypt data
+        /// to and decrypt data from this key.
+        /// </summary>
+        /// <returns>An <see cref="IAuthenticatedEncryptor"/>.</returns>
+        IAuthenticatedEncryptor CreateEncryptor();
     }
 }
--- a/src/Microsoft.AspNetCore.DataProtection/KeyManagement/KeyBase.cs
+++ b/src/Microsoft.AspNetCore.DataProtection/KeyManagement/KeyBase.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using Microsoft.AspNetCore.DataProtection.AuthenticatedEncryption;
 using Microsoft.AspNetCore.DataProtection.AuthenticatedEncryption.ConfigurationModel;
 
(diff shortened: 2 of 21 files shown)


src/Microsoft.AspNetCore.DataProtection/KeyManagement/KeyBase.cs
```diff
@@ -41,6 +52,24 @@ public IAuthenticatedEncryptorDescriptor Descriptor
...
+                    var encryptor = factory.CreateEncryptorInstance(this);
+                    if (encryptor != null)
+                    {
+                        _encryptor = encryptor;
+                        break;
+                    }
+                }
+            }
+
+            return _encryptor;
```
rynowak: Does this need to be thread safe?

src/Microsoft.AspNetCore.DataProtection/KeyManagement/IKey.cs
```diff
@@ -49,5 +49,12 @@ public interface IKey
         /// Gets the <see cref="IAuthenticatedEncryptorDescriptor"/> instance associated with this key.
         /// </summary>
         IAuthenticatedEncryptorDescriptor Descriptor { get; }
+
+        /// <summary>
+        /// Creates an <see cref="IAuthenticatedEncryptor"/> instance that can be used to encrypt data
+        /// to and decrypt data from this key.
+        /// </summary>
+        /// <returns>An <see cref="IAuthenticatedEncryptor"/>.</returns>
+        IAuthenticatedEncryptor CreateEncryptorInstance();
```
rynowak: This should be called `CreateEncryptor`

rynowak:
I'm not sure I understand the problem that's being solved here. This change basically reverts the design, now instead of being **data** the key is a **factory**

src/Microsoft.AspNetCore.DataProtection/KeyManagement/KeyBase.cs
ajaybhargavb: Yes and we already do a lock here https://github.com/aspnet/DataProtection/pull/221/files#diff-3dda400bcdb5a80421cdd26389311c63R77. The other places that we call this are just for `Debug.Assert` and we don't use those objects that are created. This has been the same since before the refactor.

ajaybhargavb:
Spoke offline. We need `CreateEncryptorInstance() on IKey` because there is no other way to associate an encryptor instance with a key.

src/Microsoft.AspNetCore.DataProtection/KeyManagement/KeyBase.cs
rynowak: You still haven't really answered this. If the invariant that we only create one is important, this code isn't correct.

ajaybhargavb: Spoke offline. This lock here was added as a perf optimization and we don't need to make changes to this.

ajaybhargavb:
🆙 📅 No longer in design

PR #230, opened 2017-05-05
Remove IHostingEnvironment dependency

Tratcher:
https://github.com/aspnet/Security/issues/1201#issuecomment-299541453
This functionality is moving into Hosting. Expect a Hosting PR shortly.

### DIFF
--- a/src/Microsoft.AspNetCore.DataProtection/DataProtectionUtilityExtensions.cs
+++ b/src/Microsoft.AspNetCore.DataProtection/DataProtectionUtilityExtensions.cs
@@ -4,7 +4,6 @@
 using System;
 using System.ComponentModel;
 using Microsoft.AspNetCore.DataProtection.Infrastructure;
-using Microsoft.AspNetCore.Hosting;
 using Microsoft.Extensions.DependencyInjection;
 
 namespace Microsoft.AspNetCore.DataProtection
@@ -37,10 +36,6 @@ namespace Microsoft.AspNetCore.DataProtection
             if (services != null)
             {
                 discriminator = services.GetService<IApplicationDiscriminator>()?.Discriminator;
-                if (discriminator == null)
-                {
-                    discriminator = services.GetService<IHostingEnvironment>()?.ContentRootPath;
-                }
             }
 
             // Remove whitespace and homogenize empty -> null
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.DataProtection/DataProtectionUtilityExtensions.cs
```diff
@@ -37,10 +36,6 @@ public static string GetApplicationUniqueIdentifier(this IServiceProvider servic
             if (services != null)
             {
                 discriminator = services.GetService<IApplicationDiscriminator>()?.Discriminator;
```
davidfowl: Make this GetRequiredService

JunTaoLuo: Hmm... Breaks a few tests. I'll merge this now and maybe fix the tests later.

PinpointTownes: @davidfowl not really a good idea. Even if this service is now registered by the hosting block, there are cases where no `IApplicationDiscriminator` will be available: when using DP in a non-ASP.NET Core app or in an OWIN/Katana app using the interop package.

PinpointTownes:
Woot, nice. This even comes with a bonus: OWIN/Katana apps using `Microsoft.Owin.Security.Interop` will no longer bring the ASP.NET Core HTTP/Hosting abstractions that were not really needed :+1:

src/Microsoft.AspNetCore.DataProtection/DataProtectionUtilityExtensions.cs
davidfowl: @PinpointTownes as long as it works with null (which it seems like it does) I guess that's fine.

PinpointTownes: This service won't be automatically registered when using DP without the hosting stack so `GetRequiredService` will throw an exception an prevent the app from loading/working correctly.

Eilon:
Yup, that was always a weird dependency.

PR #223, opened 2017-04-20
Ignore assembly version when activating DataProtection types from string name

natemcmaster:
Resolves https://github.com/aspnet/DataProtection/issues/187 by ignoring assembly version in the code the rehydrates types from strings.

Incidental fixups:

 - register activator in DI with TryAddSingleton
 - update Redis package to 0.3.0. It was building as 0.1.0, even though 0.1.1 has been released

PR #234, opened 2017-05-16
Target netstandard2.0 and add support for protecting keys with certificates on .NET Core

### LINKED ISSUE #80: Finish/verify non-windows support
There are several code paths that throw PlatformNotSupportedException on non-windows. See:
https://github.com/aspnet/DataProtection/search?utf8=%E2%9C%93&q=PlatformNotSupportedException
https://github.com/aspnet/DataProtection/search?utf8=%E2%9C%93&q=AssertPlatformIsWindows
https://github.com/aspnet/DataProtection/issues/60

We at least need to verify how functional everything is despite these gaps, like if we really can run MusicStore on coreclr on mac or linux?

@muratg

natemcmaster:
Resolves #80 

Changes:
 - Retarget Microsoft.AspNetCore.DataProtection.\* libraries to netstandard2.0. (NB: Microsoft.AspNetCore.Cryptography.\* remain netstandard1.3/net451)
 - Make available APIs that were previously only .NET Framework.
   - `IDataProtectionBuilder.ProtectKeysWithCertificate(string thumbprint)`
   - `DataProtectionProvider.Create(string applicationName, X509Certificate2 certificate)`
   - `DataProtectionProvider.Create(DirectoryInfo keyDirectory, X509Certificate2 certificate)`
   - `DataProtectionProvider.Create(DirectoryInfo keyDirectory,  Action<IDataProtectionBuilder> setupAction, X509Certificate2 certificate)`
   - `CertificateXmlEncryptor`
   - `ICertificateResolver`

Possible breaking change:
 - ~~Because `env:LOCALAPPDATA` is not normally defined on macOS/Linux, FileSystemXmlRepository would place xml key files in `~/.aspnet/DataProtection-Keys`. In .NET Core 2.0, has implemented support for `Environment.SpecialFolders.LocalApplicationData` which would map to `env:XDG_DATA_HOME` or `~/.local/share/ASP.NET/DataProtection-Keys`~~. **UPDATED**: we won't make this change.


src/Microsoft.AspNetCore.DataProtection/TypeForwardingActivator.cs
```diff
@@ -39,16 +40,14 @@ internal object CreateInstance(Type expectedBaseType, string originalTypeName, o
                 forwardedTypeName = originalTypeName.Replace(OldNamespace, CurrentNamespace);
             }
 
-#if NET46
-            if (candidate || forwardedTypeName.Contains(CurrentNamespace))
+            if (!RuntimeInformation.FrameworkDescription.StartsWith(".NET Core"))
```
pakrym: Is there a better way? I expect us to hit similar problems in other repos. Maybe we need a common helper.

natemcmaster: This is the simplest solution I've found. If you find a better one, let me know.

test/Microsoft.AspNetCore.DataProtection.Extensions.Test/DataProtectionProviderTests.cs
```diff
@@ -57,13 +57,17 @@ public void System_NoKeysDirectoryProvided_UsesDefaultKeysDirectory()
...
                 var newFileName = Assert.Single(Directory.GetFiles(keysPath));
                 var file = new FileInfo(newFileName);
                 Assert.StartsWith("key-", file.Name, StringComparison.OrdinalIgnoreCase);
                 var fileText = File.ReadAllText(file.FullName);
-                Assert.DoesNotContain("Warning: the key below is in an unencrypted form.", fileText, StringComparison.Ordinal);
-                Assert.Contains("This key is encrypted with Windows DPAPI.", fileText, StringComparison.Ordinal);
+                // On Windows, validate that it's protected using Windows DPAPI.
+                if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+                {
+                    Assert.DoesNotContain("Warning: the key below is in an unencrypted form.", fileText, StringComparison.Ordinal);
```
pakrym: `else { Assert.Contains("Warning: the key below is in an unencrypted form.", fileText, StringComparison.Ordinal); }` ?

src/Microsoft.AspNetCore.DataProtection.Extensions/Microsoft.AspNetCore.DataProtection.Extensions.csproj
```diff
@@ -4,7 +4,7 @@
 
   <PropertyGroup>
     <Description>Additional APIs for ASP.NET Core data protection.</Description>
-    <TargetFrameworks>net46;netstandard1.3</TargetFrameworks>
```
Tratcher: Should we be cross-compiling this to netstandard1.3 and netstandard2.0? This is just lightup right?

natemcmaster: It's the top of the package stack so we would also have to cross-compile Microsoft.AspNetCore.DataProtection and .Abstractions to netstandard1.3 to make that happen.

natemcmaster:
What are your thoughts on the breaking change to the location where `FileSystemXmlRepository` stores files?

src/Microsoft.AspNetCore.DataProtection.Extensions/Microsoft.AspNetCore.DataProtection.Extensions.csproj
Tratcher: .Abstractions is still 1.3. My concern is that you'll force anyone who depends on DP to move up to 2.0 even if they don't care about this new feature.

Tratcher:
Changing where the files are stored would be bad....

pakrym:
@natemcmaster why breaking change?

natemcmaster:
@pakrym in PR description.

> In .NET Core 2.0, has implemented support for Environment.SpecialFolders.LocalApplicationData which would map to env:XDG_DATA_HOME or ~/.local/share/ASP.NET/DataProtection-Keys

What if we provided an "obsolete" warning about the old file location in 2.0, and 3.0 we moved the location? 

Something like:
```
warning: Storing keys in $HOME/.aspnet/DataProtection-Keys is obsolete. In a future version, these files will no longer be loaded by DataProtection. 
         The recommended location is $HOME/.local/share/ASP.NET/DataProtection-Keys.
         You can use the new behavior now by moving the files.

    mkdir -p $HOME/.local/share/ASP.NET/
    mv $HOME/.aspnet/DataProtection-Keys $HOME/.local/share/ASP.NET/DataProtection-Keys
```

Tratcher:
@blowdart ?

src/Microsoft.AspNetCore.DataProtection.Extensions/Microsoft.AspNetCore.DataProtection.Extensions.csproj
natemcmaster: Which runtimes are in .NET Standard 1.3 but not .NET Standard 2.0 that we intend to support in 2.0 RTM?

blowdart:
Running apps which people never interact with would break.

natemcmaster:
Ok, I've updated this to preserve compatibility with 1.x. `$HOME/.aspnet` will keep precedence over `$XDG_CONFIG_HOME/ASP.NET` on Linux/macOS.

blowdart:
Please ensure the retargeting doesn't break the compat package.

src/Microsoft.AspNetCore.DataProtection.SystemWeb/Microsoft.AspNetCore.DataProtection.SystemWeb.csproj
```diff
@@ -4,7 +4,7 @@
 
   <PropertyGroup>
     <Description>A component to allow the ASP.NET Core data protection stack to work with the ASP.NET 4.x &lt;machineKey&gt; element.</Description>
-    <TargetFramework>net46</TargetFramework>
+    <TargetFramework>net461</TargetFramework>
```
natemcmaster: @blowdart is it okay to lift the compat package to require .NET Framework 4.6.1? If not, then we need to cross-compile DataProtection for net46 as netstandard2.0 requires net461.

blowdart:
What is it now? 4.5 needed?

natemcmaster:
Microsoft.AspNetCore.DataProtection.SystemWeb 1.x supported .NET Framework 4.5.1 and 2.0.0-preview1 lifted that to .NET Framework 4.6. I'm still not clear on PM requirements for desktop framework support.

pranavkm:
Are we going to continue cross-compiling https://github.com/aspnet/DataProtection/blob/dev/src/Microsoft.AspNetCore.Cryptography.Internal/Microsoft.AspNetCore.Cryptography.Internal.csproj#L7?

src/Microsoft.AspNetCore.DataProtection/TypeForwardingActivator.cs
Eilon: This looks like a no-go to me. At the very least, the check above isn't using the proper culture, but this still seems like a wrong way to check what we want here. Is there an actual reliable way to check the platform that doesn't use strings?

src/Microsoft.AspNetCore.DataProtection/TypeForwardingActivator.cs
```diff
@@ -39,16 +40,14 @@ internal object CreateInstance(Type expectedBaseType, string originalTypeName, o
...
             {
-                candidate = true;
-                forwardedTypeName = RemoveVersionFromAssemblyName(forwardedTypeName);
+                if (candidate || forwardedTypeName.Contains(CurrentNamespace))
```
Eilon: This check looks suspicious for two reasons:

1. Wrong culture used
2. To check the namespace, you want to check something more like `forwardedTypeName.StartsWith(CurrentNamespace + ".", StringComparison.Ordinal)`

... and please check all other places in DataProtection where maybe the wrong culture is being used.

src/Microsoft.AspNetCore.DataProtection/TypeForwardingActivator.cs
pranavkm: > Is there an actual reliable way to check the platform that doesn't use strings?

We could continue cross-compiling the binary. I filed an ask in CoreFx to see if they can expose a first class API: https://github.com/dotnet/corefx/issues/19927. They use FrameworkDescription in test code, but it doesn't seem reliable enough for shipping code especially given that we're tied to the `RuntimeInformation` reference and not to it's implementation.

weshaggard: What is the harm in always removing the version number from the name?

natemcmaster: When we added this code in #223, there goal was to make this fix as conservative as possible. There is no need to strip version numbers on .NET Core. 

But now, given that we have yet-to-determine a reliable way to detect .NET Core vs .NET Framework w/o cross-compiling, IMO we are better off always stripping the version number.

natemcmaster:
Blocked on https://github.com/dotnet/standard/issues/355

test/Microsoft.AspNetCore.DataProtection.Extensions.Test/DataProtectionProviderTests.cs
natemcmaster: Added.

natemcmaster:
Blocked issue was moved to https://github.com/dotnet/corefx/issues/20015.

natemcmaster:
@Eilon are we okay to merge this to unblock other repos lifting to ns2.0?

natemcmaster:
:bell: @pakrym @Eilon

build/dependencies.props
```diff
@@ -1,10 +1,10 @@
 <Project>
   <PropertyGroup>
     <AspNetCoreVersion>2.0.0-*</AspNetCoreVersion>
-    <CoreFxVersion>4.3.0</CoreFxVersion>
+    <CoreFxVersion>4.4.0-*</CoreFxVersion>
```
dougbu: Is this a required version bump? I didn't make the same change in aspnet/Localization and see @JunTaoLuo left https://github.com/aspnet/HttpAbstractions/pull/840/files#diff-30dcc9789f8ea3c7632e4f054e60c072R4 alone in his similar PR.

natemcmaster: Yes, this should be bumped. The 4.4.0 packages often include new ns2.0 runtime libraries or reference assemblies

JunTaoLuo: Ah I need to go update httpabstractions then. Yes we should be using 4.4.0. I don't think those packages were mirrored on our feeds when I started that pr and I meant to go back after the mirror builds finished.

Issue #233, opened 2017-05-16
KeyRingProvider initialization causes thread pool starvation when server starts under load


1amirjalai:
my application give me frequent 

asp.net core 1.1.2 behind load balancer iis arr3

```
HTTP Error 502.3 - Bad Gateway
The specified CGI application encountered an error and the server terminated the process.
```
i made a dump file and inspected it using windbg i encountered the following lines. it's like all thread are waiting for keymanagement 

could you please take a look at it my app crashes frequently and its a high traffic website
any help would be appreciated

```
0000001c6f67d5e8 00007ffa506f6c24 [GCFrame: 0000001c6f67d5e8] 
0000001c6f67d838 00007ffa506f6c24 [GCFrame: 0000001c6f67d838] 
0000001c6f67d878 00007ffa506f6c24 [HelperMethodFrame_1OBJ: 0000001c6f67d878] System.Threading.Monitor.ReliableEnterTimeout(System.Object, Int32, Boolean ByRef)
0000001c6f67d9a0 00007ff9bdcbd2fd Microsoft.AspNetCore.DataProtection.KeyManagement.KeyRingProvider.GetCurrentKeyRingCore(System.DateTime)
0000001c6f67da50 00007ff9bdcbc5a1 Microsoft.AspNetCore.DataProtection.KeyManagement.KeyRingBasedDataProtector.Protect(Byte[])
0000001c6f67db10 00007ff9bdcbb839 Microsoft.AspNetCore.Session.CookieProtection.Protect(Microsoft.AspNetCore.DataProtection.IDataProtector, System.String)
0000001c6f67db40 00007ff9bdcb7966 Microsoft.AspNetCore.Session.SessionMiddleware+d__9.MoveNext()
0000001c6f67dbd0 00007ff9bdcb6d45 System.Runtime.CompilerServices.AsyncTaskMethodBuilder.Start[[Microsoft.AspNetCore.Session.SessionMiddleware+d__9, Microsoft.AspNetCore.Session]](d__9 ByRef)
0000001c6f67dc30 00007ff9bdcb6cc9 Microsoft.AspNetCore.Session.SessionMiddleware.Invoke(Microsoft.AspNetCore.Http.HttpContext)
0000001c6f67dce0 00007ff9bdcb30f9 Microsoft.AspNetCore.Builder.RouterMiddleware+d__4.MoveNext()
[...]

Issue #193, opened 2016-12-14
Delay on first request due to key encryption at rest


Eilon:
_From @shravan2x on December 11, 2016 8:51_

I have an Asp.NET Core 1.1 MVC app. On every run on the app, it stalls for 3 seconds just before this is printed:

```
info: Microsoft.Extensions.DependencyInjection.DataProtectionServices[0]
      User profile is available. Using 'C:\Users\name\AppData\Local\ASP.NET\DataProtection-Keys' as key repository and Windows DPAPI to encrypt keys at rest.
```

I don't mind the log itself, but what causes the delay? Is this a bug?

_Copied from original issue: aspnet/Mvc#5613_

Eilon:
@blowdart should this bug move to the https://github.com/aspnet/DataProtection/ repo?

Eilon:
_From @blowdart on December 14, 2016 22:12_

Maybe. I'm not sure I would consider it a bug to be honest, the first request is always going to block until we get a key ring open.

blowdart:
That long a delay is a little weird though.

shravan2x:
Thanks for moving this. It always takes at least 3s. Sometimes even more.

natemcmaster:
@shravan2x can you share more code to help repro the issue? A minimal repro that isolates the issue to just DataProtection APIs is ideal. MVC adds a lot of overhead. But if that's too much work, can you share your Startup.cs file?

Btw, if your only usage is calling `.AddDataProtection()` in Startup, then I'd guess the delay is the collective time it takes to wire up services, controllers, loggers, etc., not to mention time the runtime takes to JIT all that code. It just so happens that most of the logging in an ASP.NET Core MVC app doesn't happen until after that is done, and this particular log message happens in the middle of service discovery.

shravan2x:
@natemcmaster You can find my Startup.cs [here](https://gist.github.com/shravan2x/0fe4d5a0b345eea0a939ac0b650e88a1). I have removed some instance objects unrelated to MVC or anything else.

I used to have this problem with all my projects, but only some continue to. I believe this is because I aggressively removed packages that I didn't think were important. I have never intentionally setup data protection. It probably came along with MVC or so.

I would love to compare some of my projects with+without the issue and try to corner the cause myself, but this is finals week and I won't be able to until long past the 2.0 release on May 10th =/.

natemcmaster:
[...]

PR #239, opened 2017-06-01
Add an abstraction for pre-initializing DataProtection

natemcmaster:
Companion PR https://github.com/aspnet/Hosting/pull/1096

Should reduce delay on first request caused by initializing data protection (#193) and avoid thread pool starvation when server starts under high load ( #233 ).


test/Microsoft.AspNetCore.DataProtection.Test/InitializerTests.cs
```diff
@@ -0,0 +1,33 @@
...
+using Xunit;
+
+namespace Microsoft.AspNetCore.DataProtection
+{
+    public class InitializerTests
+    {
+        [Fact]
+        public void LoadsKeyRingDuringInitialization()
+        {
+            var mock = new Mock<IKeyRingProvider>();
```
ajaybhargavb: //Arrange //Act //Assert

natemcmaster:
We decided to flip the Hosting <-> DataProtection dependency order (again) so that we don't have to do this.

See https://github.com/aspnet/DataProtection/pull/240

PR #240, opened 2017-06-01
Use IHostingEnvironment to determine application identifier

natemcmaster:
Reverses changes made in #230, which were originally made because hosting was not going to support ns2.0.

Expect a hosting PR too soon.

### DIFF
--- a/src/Microsoft.AspNetCore.DataProtection/DataProtectionServiceCollectionExtensions.cs
+++ b/src/Microsoft.AspNetCore.DataProtection/DataProtectionServiceCollectionExtensions.cs
@@ -4,6 +4,7 @@
 using System;
 using Microsoft.AspNetCore.Cryptography.Cng;
 using Microsoft.AspNetCore.DataProtection;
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.DataProtection/DataProtectionUtilityExtensions.cs
```diff
@@ -36,6 +37,11 @@ public static string GetApplicationUniqueIdentifier(this IServiceProvider servic
             if (services != null)
             {
                 discriminator = services.GetService<IApplicationDiscriminator>()?.Discriminator;
+
+                if (discriminator == null)
+                {
+                    discriminator = services.GetService<IHostingEnvironment>()?.ContentRootPath;
```
Tratcher: I liked the new implementation better. Rather than explicitly calling out to IHostingEnvironment, instead register an IApplicationDiscriminator that does it.

natemcmaster:
🆙 :date:

PR #244, opened 2017-06-05
Docker: add warning about ephermeral keys folder

### LINKED ISSUE #185: Warn about ephemeral keys when running in docker containers
Users should persist keys to an external key store (e.g. redis, named data volume) when running in docker. When this isn't configured, we should issue a warning.

One implementation idea: detect docker from `/proc/1/cgroup`

cref https://github.com/aspnet/MusicStore/issues/716

natemcmaster:
Storing keys on the file system in Docker is usually the wrong place to store things. Redis, Azure, or some other remote store are better.

When running in a Docker container (currently only Linux containers), this warning will appear whenever FileSystemXmlRepository writes a key to a folder that is _not_ volume mounted. The presumption is that volume-mounted folders is the only way to persist filesystem keys beyond the container's lifetime.
```
warn: Microsoft.AspNetCore.DataProtection.Repositories.FileSystemXmlRepository[60]
      Storing keys in a directory '/keys' that may not be persisted outside of the container. Protected data will be unavailable when container is destroyed.
```

Resolves #185


src/Microsoft.AspNetCore.DataProtection/Internal/DockerUtils.cs
```diff
@@ -0,0 +1,99 @@
...
+                    dir = dir.Parent;
+                }
+            }
+
+            return false;
+        }
+
+        private static bool IsProcessRunningInDocker()
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
```
pakrym: Try catch around everything so initialization process does not throw if something goes wrong here?

test/Microsoft.AspNetCore.DataProtection.Test/DockerUtilsTests.cs
```diff
@@ -0,0 +1,56 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.IO;
+using Microsoft.AspNetCore.DataProtection.Internal;
+using Microsoft.AspNetCore.Testing.xunit;
+using Xunit;
+
+namespace Microsoft.AspNetCore.DataProtection.Test
+{
+    public class DockerUtilsTests
```
pakrym: If functional test realistic?

natemcmaster: These lines are the actual contents of a /proc/self/mounts file with a volume mounted folder inside a microsoft/dotnet docker image.

pakrym: I don't mean these lines, I'm talking about functional test for entire feature

natemcmaster: I opened aspnet/Testing#277

src/Microsoft.AspNetCore.DataProtection/Repositories/FileSystemXmlRepository.cs
```diff
@@ -191,6 +192,12 @@ public virtual void StoreElement(XElement element, string friendlyName)
             }
 
             StoreElementCore(element, friendlyName);
+
+            if (DockerUtils.IsDocker && !DockerUtils.IsVolumeMountedFolder(Directory))
```
pakrym: Why are we logging here vs https://github.com/aspnet/DataProtection/blob/c959795a6461a9150fbabfd5c08486d05e363057/src/Microsoft.AspNetCore.DataProtection/KeyManagement/XmlKeyManager.cs#L456 . I don't think user cares about warning message if he configured directory himself. Also having it on save is a bit undiscoverable, if key already exists user won't see the warning until it expires.

natemcmaster: IMO the warning is still relevant, even if user has configured a custom directory.

Moving the warning to the class ctor seems appropriate. Will update

natemcmaster:
:up: :date:

Currently blocked on CI producing a build of aspnet/Testing with the DockerOnlyAttribute.

src/Microsoft.AspNetCore.DataProtection/Internal/DockerUtils.cs
```diff
@@ -0,0 +1,99 @@
...
+
+namespace Microsoft.AspNetCore.DataProtection.Internal
+{
+    internal static class DockerUtils
+    {
+        private static Lazy<bool> _isDocker = new Lazy<bool>(IsProcessRunningInDocker);
+
+        public static bool IsDocker => _isDocker.Value;
+
+        public static bool IsVolumeMountedFolder(DirectoryInfo directory)
```
pakrym: We are doing tons of file operations and parsing, I would prefer peoples apps not crashing if they have slightly invalid fstab. It's better to wrap this thing in try catch.

natemcmaster:
I put a try/catch around calls to dockerutils in the FileSystemXmlRepository ctor. This is currently the only place these APIs are used and should prevent the app from crashing if there are fstab formatting issues.

PR #256, opened 2017-07-06
Remove NETSTandard.Library.NETFramework

ryanbrandenburg:


ryanbrandenburg:
I'm going to make a second pass and remove `NETStandardLibraryNETFrameworkVersion` from EVERY repo. I wasn't 100% sure it was safe to kill when I made these ~30 PR's.

Issue #235, opened 2017-05-20
[1.0.x] Version-insensitive type activation on .NET Framework


natemcmaster:
Consider backporting the fix for https://github.com/aspnet/DataProtection/issues/187 to 1.x. The fix was added to 2.0.0 in this PR: https://github.com/aspnet/DataProtection/pull/223

cc @Eilon

Tratcher:
This issue impacts the ability of 2.0 and 1.1 apps to share auth cookies. https://github.com/aspnet/Release/issues/181

Eilon:
Changed this bug to be 1.0.x-specific, and logged https://github.com/aspnet/DataProtection/issues/265 to be 1.1.x-specific. That way we make sure to fix it everywhere (if we take this fix).

For future reference: please don't set any labels, the triage team will do that.

Eilon:
This patch is *approved* for 1.0.x. Please follow the patch creation instructions that were sent by email, and submit a PR to the correct branch.

Eilon:
@natemcmaster all yours.

natemcmaster:
Resolved in https://github.com/aspnet/DataProtection/pull/268. This fix will be in the 1.0.4 patch update of DataProtection (when that is released).

PR #268, opened 2017-08-25
[1.0.4] Ignore assembly version when activating DataProtection types from string name

natemcmaster:
This is a port of #223 from the 2.0.0 release. It resolves an issue caused by activating types by type name. Type.GetType enforces the assembly version. This change strips assembly version before attempting to re-create DataProtection types.

Resolves https://github.com/aspnet/DataProtection/issues/235


natemcmaster:
Automated tests pass. I'm building a repro to manually verify it.

natemcmaster:
Yup, it works. Here's a verification app for anyone else that wants to give it a go. https://gist.github.com/natemcmaster/7dfb37813a3024f228971ea89188dfde

Issue #265, opened 2017-07-28
[1.1.x] Version-insensitive type activation on .NET Framework


Eilon:
See https://github.com/aspnet/DataProtection/issues/235

Eilon:
This patch is *approved* for 1.1.x. Please follow the patch creation instructions that were sent by email, and submit a PR to the correct branch.

natemcmaster:
Resolved in https://github.com/aspnet/DataProtection/pull/269.

This fix will be available in the 1.1.3 patch of DataProtection.

PR #269, opened 2017-08-25
[1.1.3] Ignore assembly version when activating DataProtection types from string

natemcmaster:
This is a port of #223 from the 2.0.0 release. It resolves an issue caused by activating types by type name. Type.GetType enforces the assembly version. This change strips assembly version before attempting to re-create DataProtection types.

Resolves https://github.com/aspnet/DataProtection/issues/265

Manually verified: https://gist.github.com/natemcmaster/74e90b855d3d1c09aad11b683150d411


pakrym:
If it works :)

Issue #274, opened 2017-09-05
External DI Container issue


ipjohnson:
I'm having an issue with the refactoring as part of [this PR](https://github.com/aspnet/DataProtection/commit/cde3b96aa7d1e51414ab02dc2c8f22649b595da3)

The problem is for any container that allows concrete types to be automatically resolved (most allow this to happen) the wrong constructor will chosen for [KeyManagementOptionsSetup](https://github.com/aspnet/DataProtection/blob/dev/src/Microsoft.AspNetCore.DataProtection/Internal/KeyManagementOptionsSetup.cs)

This all stems from the issue raised [here](https://github.com/ipjohnson/Grace/issues/116). Normally I'd just have the user work around it but with the type being internal there isn't an easy work around. I can have the person turn off automatic registration but that's a pain and really isn't going to work for certain situations. 

What seems like the proper solution is to make the RegistryPolicyResolver parameter an interface rather than passing around the actual type. @davidfowl any thoughts?

davidfowl:
This came up here https://github.com/aspnet/DataProtection/issues/270 as well. I think we should fix it. However, we can't make any breaking changes in patches so we'd probably have to add another ctor.

/cc @muratg @Eilon @DamianEdwards

muratg:
@pakrym could you please take a look to see what's the best way to fix this without a breaking change?

davidfowl:
Actually we don't add APIs in patches either, this might need to be a 2.1 thing.

muratg:
@davidfowl Yup, it's triaged in 2.1.

davidfowl:
@muratg It might be possible to patch this since these types are internal.

Eilon:
Which internal API is it and from where are the internal APIs called? Even in patches we are careful about changing internal APIs if they're used in other product assemblies and not just tests.

davidfowl:
Here https://github.com/aspnet/DataProtection/blob/ee009982dc96171cf62ab4542aaf7c9cff984390/src/Microsoft.AspNetCore.DataProtection/Internal/KeyManagementOptionsSetup.cs#L14

PR #275, opened 2017-09-18
Make RegistryPolicyResolver an interface to fix 3d party DI

pakrym:
Fixes https://github.com/aspnet/DataProtection/issues/274

Some containers would try to activate classes even if they were not registered, switch to interface to prevent it.


src/Microsoft.AspNetCore.DataProtection/RegistryPolicyResolver.cs
```diff
@@ -14,10 +14,15 @@
 
 namespace Microsoft.AspNetCore.DataProtection
 {
+    internal interface IRegistryPolicyResolver
```
natemcmaster: Can you add a comment explaining why we have an internal only interface with only one implementation? I can easily imagine some future dev removing this because we don't have tests for the scenarios that this fixes.

PR #284, opened 2017-10-27
Update bootstrapper

ryanbrandenburg:
This Repo still needs the updated bootstrappers.


ryanbrandenburg:
Yeah, it wasn't on the list that I use to mass update repos, it should be now.

PR #298, opened 2018-02-14
Rybrande/experimental21

ryanbrandenburg:
As part of aspnet/Universe#883.


version.props
```diff
@@ -1,11 +1,15 @@
 <Project>
   <PropertyGroup>
     <VersionPrefix>2.1.0</VersionPrefix>
-    <ExperimentalProjectVersionPrefix>0.4.0</ExperimentalProjectVersionPrefix>
     <VersionSuffix>preview1</VersionSuffix>
+    <ExperimentalVersionPrefix>0.4.0</ExperimentalVersionPrefix>
```
natemcmaster: Can you leave as ExperimentalProjectVersionPrefix? This is what we are using in other repos, and it matches what our update scripts will look for.

natemcmaster: example: https://github.com/aspnet/Universe/blob/release/2.0/scripts/PatchVersionPrefix.ps1#L45

ryanbrandenburg: I went off [logging](https://github.com/aspnet/Logging/blob/dev/version.props#L5) like you said. Do we need to update logging? What's another example?

ryanbrandenburg: Ok I'll change it, but we should circle back and fix logging and anywhere else with the wrong name.

natemcmaster: Oh, I didn't realize logging was wrong too. Yeah, let's make sure we're consistent. Now I wonder what is in release/2.0. I thought we had used ExperimentalProjectVersionPrefix everywhere

src/Microsoft.AspNetCore.DataProtection.Redis/Microsoft.AspNetCore.DataProtection.Redis.csproj
```diff
@@ -4,6 +4,7 @@
     <Description>Redis storage support as key store.</Description>
     <VersionPrefix Condition="'$(ExperimentalProjectVersionPrefix)' != ''">$(ExperimentalProjectVersionPrefix)</VersionPrefix>
```
natemcmaster: We also need to flow through ExperimentalVersionSuffix since this may vary from VersionSuffix

ryanbrandenburg: `ExperimentalVersionSuffix` or `ExperimentalProjectVersionSuffix`?

ryanbrandenburg:
:up::date:

ryanbrandenburg:
:up::date:

PR #299, opened 2018-02-16
Support protecting keys with an X509Certificate even if that cert isn't in the X509Store

natemcmaster:
Fixes https://github.com/aspnet/Home/issues/2321

The default implementation of EncryptedXml doesn't support using the RSA key from X509Certificate to decrypt xml unless that cert is in the X509 CurrentUser\My or Localmachine\My store. This adds support for decrypting with the X509Certificate directly. This is useful for Linux (often Docker) scenarios, where the user already has a .pfx file, but may not have added it to X509Store.

cc @GrabYourPitchforks @blowdart @bartonjs @muratg


test/CreateTestCert.ps1
```diff
@@ -0,0 +1,14 @@
...
+
+param(
+    [Parameter(Mandatory = $true)]
+    $OutFile
+)
+
+$password = ConvertTo-SecureString -Force -AsPlainText -String "password"
+$cert = New-SelfSignedCertificate -DnsName "localhost" -CertStoreLocation Cert:\CurrentUser\My\
+Export-PfxCertificate -Cert $cert -Password $password -FilePath $OutFile
+Remove-Item "Cert:\CurrentUser\My\$($cert.Thumbprint)"
```
bartonjs: If you're netcoreapp-only you could use the CertificateRequest API and just make the test certs on the fly.  If you're both netfx and netcoreapp then this is a reasonable approach.

natemcmaster: Yeah, that's why I did this. I original wrote the tests for netcoreapp2.1 using CertificateRequest, but wanted to make sure the tests work for netfx too, so reverted to embedded a pfx file in the repo

blowdart:
Stupid question, because it's not immediately obvious - can this support multiple decryption certificates, to enable rollover? Imagine someone changes their cert, but still needs to decrypt with the previous cert.

natemcmaster:
>  can this support multiple decryption certificates, to enable rollover?

Yes! we just need to design an API so users can set multiple certs.

natemcmaster:
@blowdart  I opened https://github.com/aspnet/Home/issues/2884 to discuss further.

PR #300, opened 2018-02-20
Add UnprotectKeysWithAnyCertificate

natemcmaster:
Resolves https://github.com/aspnet/Home/issues/2884

Add an API for users to set additional certificates that can be used decrypt key data. These certificates are used in addition to the keys in the X509 store.


blowdart:
Needs a test for multiple keys. The second doesn't have to be valid, but I think a good test would be encrypt once with a pfx, abandon that protector, spin up a new one with a different pfx, protect, abandon, then start a new protector with certificate 2 as the encrypting certificate, load 1 & 2 into the decryptor, and check both decrypt. I assume 2 would already be in the list as it's the current encryptor anyway.

PR #303, opened 2018-04-02
Added Entity Framework Core backed IXmlRepository with tests and sample

dansward:


Eilon:
@dansward is there an issue associated with this PR that describes the scenario? If not, could you open a new issue at https://github.com/aspnet/Home/issues and describe the details? Thanks!

dansward:
A related issue exists. [Create SQL XmlRepository #2505](https://github.com/aspnet/Home/issues/2505), should I create one specifically for EF or will that one work?

Eilon:
@dansward I think that's probably sufficient. I'll let the experts in this space evaluate the submission. Thanks!

Tagging @muratg @blowdart to evaluate.

src/Microsoft.AspNetCore.DataProtection.EntityFrameworkCore/EntityFrameworkCoreDataProtectionBuilderExtensions.cs
```diff
@@ -0,0 +1,28 @@
+﻿using Microsoft.AspNetCore.DataProtection.EntityFrameworkCore;
+using Microsoft.AspNetCore.DataProtection.KeyManagement;
+using Microsoft.Extensions.DependencyInjection;
+using System;
+
+namespace Microsoft.AspNetCore.DataProtection
+{
+    /// <summary>
+    /// Contains Redis-specific extension methods for modifying a <see cref="IDataProtectionBuilder"/>.
```
blowdart: Change mention of Redis :)

src/Microsoft.AspNetCore.DataProtection.EntityFrameworkCore/Key.cs
```diff
@@ -0,0 +1,18 @@
+﻿namespace Microsoft.AspNetCore.DataProtection.EntityFrameworkCore
+{
+    /// <summary>
+    /// Code first model used by <see cref="EntityFrameworkCoreXmlRepository"/>.
+    /// </summary>
+    public class Key
```
blowdart: As this is never going to be typed in full by the user anywhere a more specific name would be better.

blowdart:
And adding @natemcmaster

src/Microsoft.AspNetCore.DataProtection.EntityFrameworkCore/Key.cs
dansward: DataProtectionKey? Should the DbContext class be renamed to match as well, DataProtectionKeyStore?

natemcmaster:
Hey @dansward, thanks for taking an initial stab at this. We haven't yet nailed down a design for the API and expected usage of storing keys in a SQL database. We're going to take a look at https://github.com/aspnet/Home/issues/2505 in our triage on Friday, and I'll get back to you then.

natemcmaster:
We discussed this in triage. First of all, thanks. We've been thinking of doing https://github.com/aspnet/Home/issues/2505 for a while. This was a good nudge to start the conversation again.

We're going to leave this PR open for now for two reasons.

1. This feature isn't something we can commit to shipping the 2.1.0 timeframe. We want to leave enough time for due diligence on the design and testing, and 2.1.0 is locking down soon.
2. We want to make sure the design of the API satisfies a few requirements that I don't think this currently implementation has. Those requirements are a bit vague still -- we are going to have a conversation with the EF Core team first -- but when we have them, I'll list those requirements in https://github.com/aspnet/Home/issues/2505. If you're open to working with us on that design and willing to update this PR to meet those, we may be able to take this PR as the starting point for the feature.

Thanks,
Nate

dansward:
@natemcmaster Sounds good. I'll wait for the update to [aspnet/Home#2505](https://github.com/aspnet/Home/issues/2505).

ekarlso:
Any way I can get migrations working with this?

ekarlso:
Anyone wanna play in on this? @dansward

dansward:
Will put together an example or fix it if I can't as soon as I have a few minutes. But, I'd try putting migrations in your assembly first.
________________________________

natemcmaster:
Hey @dansward, just wanted to let you know I haven't forgotten about this. Thanks for being patient. Our master branch is open for 3.0 work now. DataProtection is one of the dozens of projects I have to find time to support, so I appreciate your help.

What are your thoughts on this comment?  https://github.com/aspnet/Home/issues/2505#issuecomment-403146056. I think @ekarlso's comment pointed out one of the things we should discuss more - how can we help users create/manage the database to back this repository?

natemcmaster:
Just tossing out an idea. One way I thought we might implement is by providing an API like this:
```c#
    public static class EtityFrameworkDataProtectionExtensions
    {
        public static IDataProtectionBuilder PersistKeysToDbContext<TContext>(this IDataProtectionBuilder builder)
            where TContext: DbContext, IDataProtectionKeysDbSet
        {
            /// implementation ...
            return builder;
        }
    }

    public interface IDataProtectionKeysDbSet
    {
        DbSet<DataProtectionKeys> DataProtectionKeys { get; }
    }

    public class DataProtectionKeys
    {
        public long Id { get; set; }
        public string Content { get; set; }
    }
```

I would have to play with an implementation, but my thinking here is that this gives devs the option to use their existing DbContext type and just add a new table, but also gives them flexibility to keep keys in a separate DbContext if they wish. 

Also, we can create the DbContext instance through dependency injection, so devs only have to configure their connection strings in EF Core once

dansward:
@natemcmaster, that worked really well. Let me know if this meets the requirements.

src/Microsoft.AspNetCore.DataProtection.EntityFrameworkCore/EntityFrameworkCoreDataProtectionExtensions.cs
```diff
@@ -0,0 +1,28 @@
...
+        /// </summary>
+        /// <typeparam name="TContext">The EntityFrameworkCore <see cref="DbContext"/> type used to access instances of <see cref="DataProtectionKey"/></typeparam>
+        /// <param name="builder">The <see cref="IDataProtectionBuilder"/> instance to modify</param>
+        /// <returns>The value <paramref name="builder"/>.</returns>
+        public static IDataProtectionBuilder PersistKeysToDbContext<TContext>(this IDataProtectionBuilder builder)
+            where TContext : DbContext, IDataProtectionKeyContext
+        {
+            builder.Services.Configure<KeyManagementOptions>(options =>
+            {
+                options.XmlRepository = new EntityFrameworkCoreXmlRepository<TContext>(builder.Services.BuildServiceProvider().GetRequiredService<TContext>());
```
natemcmaster: I think this may cause problems with scope. Typically, DbContext is a scoped service, but this would required a DbContext to be a singleton.

src/Microsoft.AspNetCore.DataProtection.EntityFrameworkCore/EntityFrameworkCoreXmlRepository.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        /// <summary>
+        /// Creates a new instance of the <see cref="EntityFrameworkCoreXmlRepository{TContext}"/>.
+        /// </summary>
+        /// <param name="context">The context used to store instances of <see cref="DataProtectionKey"/></param>
+        public EntityFrameworkCoreXmlRepository(TContext context) => _context = context ?? throw new ArgumentNullException(nameof(context));
+
+        /// <inheritdoc />
+        public IReadOnlyCollection<XElement> GetAllElements()
+            => _context.DataProtectionKeys.Select(key => XElement.Parse(key.Xml)).ToList().AsReadOnly();
```
natemcmaster: Let's be gentle and use TryParse instead. If the database has malformed xml, we should skip loading that key and log a warning.

Also, let's make this a notracking query. We don't need to make changes to the entities.

src/Microsoft.AspNetCore.DataProtection.EntityFrameworkCore/EntityFrameworkCoreXmlRepository.cs
```diff
@@ -0,0 +1,44 @@
...
+    public class EntityFrameworkCoreXmlRepository<TContext> : IXmlRepository
+        where TContext : DbContext, IDataProtectionKeyContext
+    {
+        private readonly TContext _context;
+
+        /// <summary>
+        /// Creates a new instance of the <see cref="EntityFrameworkCoreXmlRepository{TContext}"/>.
+        /// </summary>
+        /// <param name="context">The context used to store instances of <see cref="DataProtectionKey"/></param>
+        public EntityFrameworkCoreXmlRepository(TContext context) => _context = context ?? throw new ArgumentNullException(nameof(context));
```
natemcmaster: Instead of taking an instance of `TContext`, we should take a service which can construct a new instance of the DbContext. In earlier versions of EFCore, there was an `IDbContextFactory<TContext>` service you could use. I don't remember if that is still around in EF Core 2.0 (cc @divega).

Also, EntityFrameworkCoreXmlRepository is a singleton, but each time it invokes an interaction with the database, we should create a new instance of DbContext.

src/Microsoft.AspNetCore.DataProtection.EntityFrameworkCore/EntityFrameworkCoreXmlRepository.cs
```diff
@@ -0,0 +1,44 @@
...
+            {
+                FriendlyName = friendlyName,
+                Xml = element.ToString(SaveOptions.DisableFormatting)
+            };
+            var oldKey = _context.DataProtectionKeys.Find(friendlyName);
+            if (oldKey == null)
+            {
+                _context.DataProtectionKeys.Add(newKey);
+            }
+            _context.SaveChanges();
```
natemcmaster: Can you add a logging statement to this method before you call SaveChanges? It only needs to be LogLevel.Debug that says something like "Saving key {friendlyName} to {dbcontextname}".

src/Microsoft.AspNetCore.DataProtection.EntityFrameworkCore/IDataProtectionKeyContext.cs
```diff
@@ -0,0 +1,13 @@
+﻿using Microsoft.EntityFrameworkCore;
```
natemcmaster: Formatting nit: can you apply the standard file header to all .cs files?

test/Microsoft.AspNetCore.DataProtection.EntityFrameworkCore.Test/EntityFrameworkCoreDataProtectionBuilderExtensionsTests.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+        [Fact]
+        public void PersistKeysToEntityFrameworkCore_UsesEntityFrameworkCoreXmlRepository()
+        {
+            var serviceCollection = new ServiceCollection();
+            serviceCollection
+                .AddDbContext<DataProtectionKeyContext>()
+                .AddDataProtection()
+                .PersistKeysToDbContext<DataProtectionKeyContext>();
+            var serviceProvider = serviceCollection.BuildServiceProvider();
```
natemcmaster: Can you change this to `.BuildServiceProvider(validateScopes: true)`? This typically helps us find issues in the CI services.

src/Microsoft.AspNetCore.DataProtection.EntityFrameworkCore/EntityFrameworkCoreXmlRepository.cs
bricelam: Don't use `IDbContextFactory`. It was only for design-time. We renamed it in 2.0 to reflect that.

src/Microsoft.AspNetCore.DataProtection.EntityFrameworkCore/IDataProtectionKeyContext.cs
```diff
@@ -0,0 +1,13 @@
...
+
+namespace Microsoft.AspNetCore.DataProtection.EntityFrameworkCore
+{
+    /// <summary>
+    /// Interface used to store instances of <see cref="DataProtectionKey"/> in a <see cref="DbContext"/>
+    /// </summary>
+    public interface IDataProtectionKeyContext
+    {
+        /// <inheritdoc />
+        DbSet<DataProtectionKey> DataProtectionKeys { get; }
```
bricelam: Requiring a `DbSet<DataProtectionKey>` on the `DbContext` seems a bit rigid. Could just use `DbContext.Set<DataProtectionKey>()` instead.

src/Microsoft.AspNetCore.DataProtection.EntityFrameworkCore/IDataProtectionKeyContext.cs
```diff
@@ -0,0 +1,13 @@
+﻿using Microsoft.EntityFrameworkCore;
+
+namespace Microsoft.AspNetCore.DataProtection.EntityFrameworkCore
+{
+    /// <summary>
+    /// Interface used to store instances of <see cref="DataProtectionKey"/> in a <see cref="DbContext"/>
+    /// </summary>
+    public interface IDataProtectionKeyContext
+    {
+        /// <inheritdoc />
```
bricelam: Inherit docs from where? 😉

src/Microsoft.AspNetCore.DataProtection.EntityFrameworkCore/EntityFrameworkCoreXmlRepository.cs
bricelam: When are keys retrieved? Once during app start? During a request?

If it's just once, just creating a scope here might be best.

dansward:
@natemcmaster @bricelam 
I made most of the changes requested. I did not remove the interface but am using DbContext.Set<DataProtectionKey> in the repository.

I'm not sure how to fix the conflicts. I hate to waste your time but any help would be appreciated.

natemcmaster:
Thanks for doing the initial legwork to get this going @dansward. I've pushed a commit to your branch to resolve the conflicts. I'm going to merge this in to a staging branch while I do a few more tweaks. Assuming I don't run into other issues, this will be in the 2.2 Preview 2 release.

PR #310, opened 2018-06-07
Adding VSTS file

ryanbrandenburg:


natemcmaster:
We probably need to also update the build definitions to use the new file path, ci-public.yml.

ryanbrandenburg:
Yeah, I was in the middle of doing that (it fails if the file doesn't exist in the default branch, so I couldn't do it before) when the VSTS UI started hanging on all the builds.

natemcmaster:
macOS tests fail with
```
Interop+AppleCrypto+AppleCommonCryptoCryptographicException : A default keychain could not be found.
```
This happens when the process running the tests is launched from a non-interactive session. This is something we can't workaround, unfortunately. We may have to skip those tests altogether.

ryanbrandenburg:
It looks like you might be the SME for DataProtection, if we can't run this test in VSTS is that going to be a problem for us? Can we abstract this test in a different way such that it can run and still cover its area?

natemcmaster:
So, something on the VSTS macOS agents is goofy. There is no default keychain set. I tried to add a test fixture to create a temporary default keychain, but it seems something on the agent is blocking this action.

```
2018-06-08T19:43:57.2369510Z   log: current default keychain = ''
2018-06-08T19:43:57.2470640Z   log: setting default keychain to = '/Users/vsts/agent/2.134.2/work/1/s/test/Microsoft.AspNetCore.DataProtection.Extensions.Test/bin/Release/netcoreapp2.2/xudgo2c5.z14.keychain-db'
2018-06-08T19:43:57.3180650Z   log: new default key chain = ''
```

PR #314, opened 2018-06-28
Unprotect key material with the local cache of certificates before checking the cert store

natemcmaster:
Resolves https://github.com/aspnet/Home/issues/3218

In some cases, private keys for certificates is not completely available. When attempting to decrypt key material, this can cause 'CryptographicException: Keyset does not exist'.

This changes the order in which decryption looks up private keys. It will first use certificates provided explicitly to the API for unprotecting keys, and then falling back to the cert store.

cc @blowdart


natemcmaster:
@guardrex I think the only thing to add is that starting in .NET Core 2.1, you can provide a X509Certificate2 object (such as a cert loaded from a file) to Protect/Unprotect keys at rest. This feature was added in https://github.com/aspnet/DataProtection/pull/299.

This PR is a fix to that feature, which helps resolve conflicts when you provide an X509Certificate2 object _and_ the cert is also in the X509Store. 

Sample of how to use this feature:

```c#
sc.AddDataProtection().ProtectKeysWithCertificate(new X509Certificate2("mycertfile.pfx", "securecertpassword"))
```

Also, we added `UnprotectKeysWithCertificates` for users who may want to rotate certs but still need to decrypt keys at rest.

```c#
sc.AddDataProtection().UnprotectKeysWithCertificates(new X509Certificate2("old_expired_cert1.pfx", "securecertpassword"), new X509Certificate2("old_expired_cert3.pfx", "securecertpassword")) ;
```

natemcmaster:
@pakrym can you take a look at this PR since you reviewed #299  as well?

natemcmaster:
:bell:

test/Microsoft.AspNetCore.DataProtection.Extensions.Test/DataProtectionProviderTests.cs
```diff
@@ -253,7 +302,7 @@ private static void WithUniqueTempDirectory(Action<DirectoryInfo> testCode)
                 // clean up when test is done
                 if (dirInfo.Exists)
                 {
-                    dirInfo.Delete(recursive: true);
+                    // dirInfo.Delete(recursive: true);
```
pakrym: undo

natemcmaster: good catch, thanks.

PR #320, opened 2018-08-10
Change versioning of Microsoft.AspNetCore.DataProtection.Redis to 2.2

natemcmaster:
Part of https://github.com/aspnet/Home/issues/3379


natemcmaster:
@pakrym can you take a quick look? Per @blowdart's request, this makes the package "supported" by changing its versioning to 2.2.0 instead of 0.5.0

PR #322, opened 2018-08-28
Resolve continually flaky test in 2.1 by making it possible to mock the folder used for key storage

natemcmaster:
Create an internal abstraction for finding the default directories for key storage. This allows us to run tests without squashing on keys on the developer machine. It also allows us to isolate test runs from reach other.

Fixes aspnet/Home#3034


src/Microsoft.AspNetCore.DataProtection.Extensions/DataProtectionProvider.cs
```diff
@@ -150,7 +150,7 @@ public static IDataProtectionProvider Create(string applicationName, X509Certifi
             return CreateProvider(keyDirectory, setupAction, certificate);
         }
 
-        private static IDataProtectionProvider CreateProvider(
+        internal static IDataProtectionProvider CreateProvider(
```
muratg: Pinging @Eilon to get his opinion on private to internal change here.

Eilon: Eh, this is not a test-only change. I'd rather not take this through.

Eilon: I'd rather even do an `#ifdef` or something to avoid any possible change to the DLL (even if we don't ship the DLL).

natemcmaster:
I can't fix the test without making a change to allow mocking a static method, so what do you propose we do?

Eilon:
I just posted a comment to the issue thread: can we disable parallelization of these tests?

natemcmaster:
The flakiness is not an issue with parallelization. It is a problem with bad machine state.

Eilon:
If we disable parallelization can we have cleanup code that runs before/after each affected test so that they're guaranteed not to clobber over each other?

natemcmaster:
It's not about parallelization. The global machine folder may be there for a number of reasons which disabling parallelization does not solve. when the global machine folder exists, the test attempts to avoid clobbering it by moving it temporarily. But moving it is what is causing problems. The location of the global machine folder is uninteresting in this test, which is why the fix for the flakiness is to use a different folder.

src/Microsoft.AspNetCore.DataProtection.Extensions/DataProtectionProvider.cs
natemcmaster: Can someone explain the risk you're worried about? I'm having a hard time thinking of a realistic way that this would break someone.

Eilon:
What risk is there in disabling these tests for the patch?

natemcmaster:
> What risk is there in disabling these tests for the patch?

We lose regression coverage. If you'd rather risk that, then let's just get rid of the test.

Eilon:
I think if we don't have a reasonable way of de-flakifying the test without product changes then we should disable the test.

